PR #6, opened 2022-05-09
Adding more event types

adamribaudo-velir:
Added: 
- video complete
- video start
- first visit
- user engagement
- scroll


dgitis (approved):

dgitis (commented):
-- TODO: Unclear why there are first_visit events firing when the ga_session_number is >1. This might cause confusion.

The first_visit and ga_session_number are set in different cookies. A first_visit with ga_session_number > 1 happens when the cookie with client_id is deleted and then set anew without deleting the session counter.

You could use this to create a new event reset_client_id_first_visit and replace the first_visit, but that might be confusing when others use these models.

dgitis (commented):
The video_start and video_complete events have the same parameters. I can think of some decent reasons for keeping these video tables separate, like consistency, for example, but I'd also consider combining the video events into a single table.

dgitis (approved):

PR #8, opened 2022-05-20
Custom parameters

dgitis:
Add custom parameters macro, conditional logic to events staging tables to create custom parameters, and added commented instructions for creating them in the dbt_project.yml file.


dbt_project.yml
```diff
@@ -13,10 +13,19 @@ clean-targets:         # directories to be removed by `dbt clean`
   - "target"
   - "dbt_packages"
 
-#vars:
-#    start_date: "20210120" # Defines the earliest GA4 _TABLE_SUFFIX to load into base events model. 
-#    project: "bigquery-public-data"
-#    dataset: "ga4_obfuscated_sample_ecommerce"
+profile: obfuscated_test_db
```
adamribaudo-velir, line 16: @dgitis I don't think we need/want to define a profile or vars in the package itself. They should be defined in the project that loads the package.

.gitignore
```diff
@@ -2,3 +2,6 @@
 target/
 dbt_packages/
 logs/
+dbt_project.yml
```
adamribaudo-velir, line 5: @dgitis I don't think we want to ignore the dbt_project.yml or packages.yml .

dbt_project.yml may have useful materialization defaults. packages.yml may define may define dependencies for our package

models/staging/ga4/stg_ga4__ecommerce.sql
```diff
@@ -0,0 +1,15 @@
+select
```
adamribaudo-velir, line 1: @dgitis Can you source this from `stg_ga4__events`? That's where we define event & session keys which will serve as primary keys for events and sessions and may be useful during later transformations.

.gitignore
dgitis: Definitely right about this. I'll fix it.

dbt_project.yml
dgitis: I wasn't using this. It's been in there from the beginning.

models/staging/ga4/stg_ga4__ecommerce.sql
dgitis: As requested, this will be moved to its own pull request. I'll resolve it here.

adamribaudo-velir:
I cleaned this up a bit based on my comments. However, I can't actually run `dbt run` without those variables set. I get:

`Required var 'page_view' not found in config:`

Are you sure that the `{% if var("page_view") %}` condition works the way we think it does? We'll need to make sure the project compiles even if certain events have no custom parameters.

adamribaudo-velir:
@dgitis figured this out. You need:

`{% if var("page_view", "none") != "none" %}` 

If the variable isn't set, it will be set to none. Then we can check to see if the variable == `none` . This compiles. 

I think my last comment ( I promise! ) is that we should name these variables something that indicates they are related to customer parameters. Like `page_view_custom_parameters` and `scroll_customer_parameters`. Just naming a variable `page_view` is very ambiguous.

PR #12, opened 2022-05-28
User properties

adamribaudo:
## Description & motivation
GA4 admins are able to assign any event param as a 'user dimension' in the GA4 interface. This PR provides equivalent functionality from within BigQuery by assigning event params as user-scoped dimensions. A new model, `stg_ga4__user_properties` finds the latest instance of an event parameter for each `client_id` 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,20 @@ vars:
           - name: "country_code"
             value_type: "int_value"
 ```
+### Using User Properties
+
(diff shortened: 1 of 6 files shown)


adamribaudo-velir:
Accidentally opened PR under my personal account rather than work account! woops.

@dgitis, can you review?

dgitis (approved):
I don't think we are done with user properties, but this is a great start.

adamribaudo-velir:
Realized today that I'm taking the latest value from the latest event, but I should be taking the latest event with a non-null value for the event parameter in question. TODO!

PR #14, opened 2022-05-29
Staging base static partition

dgitis:
## Description & motivation
Moved static partition into base_ga4__events.sql file and created config parameters to select static vs dynamic partitions.

Historically, our tests are just the native tests for value types which this PR does not change. It could be unit tested.

## Checklist


adamribaudo-velir:
Just left a message in Slack, but I'm not sure static partitioning adds any value. I thought it would avoid DBT needing to calculate the max partition, but it does it anyway. And BQ data is never updated after it's written, so there shouldn't be a need to reprocess data after it's written. 

I'm open to ideas, but I'd like to hear what anyone gains by using static partitioning before merging.

dgitis:
If I understand https://discourse.getdbt.com/t/bigquery-dbt-incremental-changes/982#incremental-strategies-4 correctly, using a static incremental strategy with partition_to_replace is more performant on bigger datasets while the dynamic dbt_max_partition is better for smaller datasets.

Based on this, I minimally need to switch the code to use partion_to_replace since dbt enables some BQ optimizations in the background when that config is set.

One thing I'd like to consider, given BQ's generous free tier, do we want to only use the static method and reduce the amount of configuration required? Small projects will probably stay in the free tier using the sub-optimal at small-scale static incremental strategy. Large projects are the ones where these settings will matter most, so why let people mess up their settings and incur charges needlessly?

dgitis:
I'm currently chasing down a bug where I'm getting a partition_on is not a timestamp error despite having declared the partition_on field to be a date type. I need to see if I can reproduce on dbt Cloud.

dgitis:
I can't add tests until the unit testing framework gets merged. I have this working locally and in dbt Cloud where it is running twice daily.

adamribaudo-velir:
@dgitis unit testing work is merged to main! thanks for your patience. Want to retest?

adamribaudo-velir:
@dgitis I think there may be a bug. I just ran a full refresh, deleted the last 10 days of data, then ran the static partition with 

      use_static_partition: true
      static_partition_lower_bound: 2

I ran this on 6/16 but it only pulled data from 6/14. Wouldn't users expect data to be pulled from 6/15 and 6/14? Can you check?

adamribaudo-velir:
I think I figured it out. Your code here will choose 'current_date' and '2 days ago' but nothin gin between:

{% set partitions_to_replace = [
    'current_date',
    'date_sub(current_date, interval ' + var('static_partition_lower_bound')|string + ' day)'
    ] %}

If someone asked for 10 days of data, you would need a `for` loop that would list out 10 days. Right now you just output the first and last date.

dgitis:
I tried to change that date_sub to use between in an earlier revision but every document everywhere used in so I trusted that it worked. I'll get to it. And, with the test suite added, I can add tests too.

adamribaudo-velir:
I'm not 100% sure a test for this will work because you need it to run incrementally but the default policy within DBT is to run a full refresh if the model doesn't yet exist in the destination. So I wouldn't let the unit test hold back your work if that becomes an issue.

PR #15, opened 2022-05-30
Added unit testing framework and unit tests for 2 models. 

adamribaudo-velir:
## Description & motivation

The only way to confirm that our package works as we make changes it to create unit tests for models that mock the input data and compare the expected results to the actual results. 

There have been some discussion about using `pytest` to achieve this result, which is the same method used by adapter maintainers. See https://github.com/dbt-labs/dbt-core/discussions/4455#discussioncomment-2766503

There is a readme file in the `integration_tests` folder with instructions for running the tests. 

## Checklist


dgitis:
For the test data, what if we copy a couple days worth of data from Google's GA4 sample dataset and use that as the source data for testing? The tests would take longer to run, but we wouldn't need to worry about mocking nested data and it probably wouldn't cost us anything.

dgitis (changes requested):
Modify the README.md and this will be ready to merge into main.

PR #20, opened 2022-07-04
Prefixed macros with ga4. to simplify replacing models in local project

dgitis:
## Description & motivation
When using package macros outside the package, you need to prefix the macro with the package name. In our case, `ga4.unnest_key( ...`. 

The macros work the same way with or without the prefix within the package so this isn't strictly necessary, but it makes it faster and easier to copy and replace a model.

We should make it a coding standard to prefix all of our macros with the package name going forward.

## Checklist


adamribaudo-velir:
Thanks for the tip. Didn't know this.

PR #21, opened 2022-07-06
Ecommerce

dgitis:
## Description & motivation
This is not ready to go live. I will be testing with proper ecommerce data on a client site.

I need to update documentation too.

I've flattened the ecommerce struct in the purchase and refund tags. I'm not sure that I understand the purpose of this field.

I've flattened the items object in the ecommerce events that should only have one item in them and created a separate table with all ecommerce items in them for other cases.

I created an ecommerce flag to disable ecommerce tracking.

## Checklist


dgitis:
I'm not sure how to handle documenting and testing models that might not be used. The plan for these ecommerce updates is to apply them to all recommended events. I'm going to leave that open for debate.

models/staging/ga4/stg_ga4__items.sql
```diff
@@ -0,0 +1,41 @@
+{{
+    config(
+        enabled = false,
+    )
+}}
+
+ with items_with_params as (
+    select
+        event_key,
```
adamribaudo-velir, line 9: I think it would be useful to keep the event_name in the select clause here. What do you think? If someone goes here looking for purchased items, they'll want to set a where clause on `event_name = 'purchase'`

adamribaudo-velir:
This looks good. I don't see an `ecommerce` variable flag, but I see that you set `stg_ga4__items` to disabled by default, so maybe you were thinking the end-user would update their `dbt_project.yml` to enable it if necessary? That sounds fine, but we should update the readme to cover that (Right now we only talk about the recommended events)

adamribaudo-velir (approved):

adamribaudo-velir:
@dgitis looks good! I just renamed 1 model from `items` to `event_items` to better clarify that the model contains items associated with events rather than a distinct list of items.

PR #25, opened 2022-07-10
Introduce user_key which pulls from user_id or user_pseudo_id depending on which is available

adamribaudo-velir:
## Description & motivation
When building user-level models and dimensions, it's helpful to join on a `user_id` if one is available and to fall back to `user_pseudo_id` if necessary. 

Also, if user privacy settings cause each of those values to be null, their data should be removed from user-level models such as `dim_users`.

This PR adds a new `user_key` field that should be used for joins and aggregations at the user level. This abstraction will make engineer lives easier by allowing them to work with 1 field rather than continually checking for both `user_id` and `user_pseudo_id`

## Checklist


dgitis:
We should think about setting a client_key which is always the md5 value of the client ID. This will simplify debugging. I also think that some people will appreciate letting them choose whether to query by user_key or client_key.

There might also be a use for average number of clients per user which this would enable. Can we get these data from the GA4 interface to compare Google's enhanced matching to default.

dgitis:
I'm going to approve this and let you decide whether the client_key is worth adding.

dgitis (approved):
Possibly add a client_key per my comments.

adamribaudo-velir:
I'm not clear on what value an md5 of the user_psuedo_id has over the user_pseudo_id itself. It's already psuedo anonymous so hashing it doesn't obfuscate anything further. 

We can talk about it. I'll merge for now.

PR #29, opened 2022-07-12
Basic streaming support

dgitis:
## Description & motivation
The streaming export can be used on its own. It's particularly useful for sites that hit the 1 million daily event limit of the batch export. 

These particular changes will be used as the base for some high-volume site optimizations that will involve some major changes to the package. However, the effect of this portion is minor.

This PR lets users choose to use the events_* or events_intraday_* tables as the source of their data.

Most people set the GA4 BigQuery export Frequency to Batch and Streaming. In these cases, the events_intraday_* partitions get deleted when the events_* partitions are added for those same days. As a result, most people think the events_intraday_* table is just for temporary data. 

This is wrong.

If you only set the Frequency to Streaming (as you would do if you expected to exceed the 1 million daily event limit), then no events_* table is created and the events_intraday_* table is the only table in the export and its partitions are never deleted (presuming that you have billing set up).

This PR supports this basic use case.

## Checklist


dgitis:
I replaced the include_intraday_events parameter with a frequency parameter because it seemed less confusing to match the language in GA4 rather than have one setting to control whether events_ or events_intraday_ served as the base table and another setting to control whether to append intraday tables for sites using the events_table.

I preserved the include_intraday_events variable in the code for backwards compatibility and just removed it from the documentation.

I've tested these changes on sites using both events_ and events_intraday_ tables as the base.

adamribaudo-velir:
@dgitis I know we chatted about this over Slack a while back, but can you provide a bit more documentation in the PR about what 'streaming support' means, what these changes do, and link to any relevant documentation from Google?

dgitis:
Let me know if you need any more information about the purpose of this pull request. I updated the description.

adamribaudo-velir:
@dgitis this mostly looks good but I think you forgot to replace the `include_intraday_events` variable from the `stg_ga4__events` model and update the logic to union the streaming table in the daily+streaming option. I think that's the only case where a union is necessary.

dgitis:
Okay, I cleaned a few things up as well. 

I've tested various combinations of frequency and include_intraday_events. Here are the results:

| frequency | intraday true |  intraday false |  intraday not set |
|---|---|---|---|
| streaming |  no union |  no union  | no union |
| daily |  union intraday  |  no union   | no union |
| daily+streaming | union intraday |  union intraday | union intraday |

adamribaudo-velir:
@dgitis I thought you were going to remove the `include_intraday_events` variable completely. I feel like users have enough options with the 3 settings you've defined. For example, if they want just the daily events they can set frequency to `daily`. 

Any issues with removing `include_intaday_events`?

dgitis:
I removed it from the documentation but left the code in so that upgrades
work without setting the frequency variable.

I'm fine with removing it entirely. We'll need to document the upgrade
process on the next major release.


*Damon Gudaitis*
www.caretjuice.com

adamribaudo-velir (approved):

Issue #31, opened 2022-07-18
BUG?! base_ga4__events.sql  --> Dynamic vs Static partitions IF clause


vibhorj:
**Context:** base_ga4__events.sql -> [Line #1](https://github.com/Velir/dbt-ga4/blob/105cf51d3fd4b7e1d8a0557298c7dc248e4bdd0f/models/staging/ga4/base/base_ga4__events.sql#L1) & [Line #58 ](https://github.com/Velir/dbt-ga4/blob/105cf51d3fd4b7e1d8a0557298c7dc248e4bdd0f/models/staging/ga4/base/base_ga4__events.sql#L58)

**Issues:** the conditional IF statement is not executing as expected
`{% if var('static_incremental_days', false ) ==  true %}`

**Impact**: even though I have set a variable `static_incremental_days: 3` in **dbt_project.yml** file, the compiled code still rendering statements meant for [dynamic-partitions ](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#dynamic-partitions) and not  [static-partitions ](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#static-partitions) as expected

@adamribaudo I can fix, create PR to merge if you agree ?

dgitis:
I noticed a lot of weirdness related to this feature a couple of days ago. If you know a fix, please submit it because I'm a little confused as to why it's not working. I noticed this while working on the streaming branch.

I've also posted my issue in the dbt Slack, but haven't gotten an answer there https://getdbt.slack.com/archives/C99SNSRTK/p1657911399355319.

dgitis:
Actually, I have this fixed in the streaming PR request. My issue in testing seems to be from having multiple profiles on my test machine and somehow my test is getting the wrong data creating what appears to be an error.

vibhorj:
@dgitis thank you for sharing.
yes, its the same things I notice - i have created the PR https://github.com/Velir/dbt-ga4/pull/32

It seems Jinja doe snot recognize non-null value as true

vibhorj:
@dgitis - sorry I did not want to interfere with other changes you might have made for streaming support, therefore I decided to create this PR and limit to this bug fix. thanks for co-operation.

adamribaudo-velir:
Great find! Mistake on my part: I meant to look for `!= false` rather than `== true`

PR #32, opened 2022-07-19
fixed in base_ga4__events.sql - BUG #31

vibhorj:
## Description & motivation
As discussed in Bug #31 
It seems Jinja does not recognize non-null value as true

## Checklist
- [yes ] I have verified that these changes work locally


adamribaudo-velir (approved):

adamribaudo-velir:
Great catch. Thanks!

PR #37, opened 2022-07-19
Update static incremental load to use table suffix

adamribaudo-velir:
## Description & motivation

At some point

## Checklist
- [ X] I have verified that these changes work locally
- [ X] I have run `dbt test` and `python -m pytest .` to validate exists tests


dgitis (approved):
I'd actually caught this in the streaming branch.

PR #38, opened 2022-07-22
Pages Mart

dgitis:
## Description & motivation
Page reporting is standard for every site and should be a core mart in the package.

The fct_ga4__pages mart groups page data by page_location, page_title, event_time_dt, and hour.

It includes a number common metrics that are easily derived from the page_view events table. The value comes in reporting conversions on page which is not as simple to report on.

page_title should strictly be in its own dim_ga4__pages model. Implementing this model would simplify adding page dimensions like author, reading difficulty and the like. However, thought needs to be put in to how to handle changes to page titles in any dim_ga4__pages model. Including the page_title in the fct_ table seems like a reasonable sacrifice to expediency.

## Checklist
- [ y/n] I have run `dbt test` and `python -m pytest .` to validate exists tests
I'm having problems with the .env file. I need some time to sort those out before running pytest.


models/staging/ga4/stg_ga4__items.sql
```diff
@@ -0,0 +1,42 @@
+{{
```
adamribaudo-velir, line 1: @dgitis if you look at `main`, this was renamed to `stg_ga4_event_items.sql` to better describe that this model contains event items (pk is event_key), as opposed to a model of items (pk is item_id).

Can you delete and refactor?

models/marts/core/fct_ga4__pages.sql
```diff
@@ -0,0 +1,45 @@
+with page_view as (
+    select
+        event_date_dt,
+        extract( hour from (select  timestamp_micros(event_timestamp))) as hour,
+        page_location,  -- does not include query string; disable this model, copy this to your project and switch to original_page_location to segment by page_location with query string
```
adamribaudo-velir, line 5: responding to your comment on this line... `page_location` does include the query string. The user can optionally strip specific query parameters from the query string using the `query_parameter_exclusions` var, but again, that only strips specific query string params. 

So I'm not sure the comment is needed, in fact you may want to say "-- page_location includes any query string params that were not filtered out by the query_parameter_exclusions variable"

models/marts/core/fct_ga4__pages.sql
```diff
@@ -0,0 +1,45 @@
+with page_view as (
+    select
+        event_date_dt,
+        extract( hour from (select  timestamp_micros(event_timestamp))) as hour,
+        page_location,  -- includes query string parameters not listed in query_parameter_exclusions variable
+        concat( cast(event_date_dt as string), cast(extract( hour from (select  timestamp_micros(event_timestamp))) as string), page_location ) as page_key,
+        page_title,  -- would like to move this to dim_ga4__pages but need to think how to handle page_title changing over time
+        count(event_name) as visits,
```
adamribaudo-velir, line 8: @dgitis ok to rename to "pageviews"? Visits is often associated with sessions. But this is a count of pageviews ,yes?

dgitis: I used page_views as it seems more consistent with other column names like new_users.

adamribaudo-velir (approved):

PR #53, opened 2022-08-23
Flattened records

dgitis:
Tableau can't query BigQuery records. 

In order to address this difficulty, this PR flattens geo, traffic_source, app_info, device, user_ltv, and privacy_info records.

Event params and user_properties were left nested as we already have more flexible ways of unnesting these records while ecommerce and items are already addressed manually.

## Checklist


willbryant:
Metabase too FWIW. I've been using another view layer on top, would be nice not to need to.

adamribaudo-velir (approved):

PR #58, opened 2022-08-24
Support derived_session_properties (#54)

### LINKED ISSUE #54: Derived session properties
`derived_user_properties` is excellent and exactly what I needed for some of my properties that once set don't really change.

But we also have some event params that will most likely be different for each session, and putting them in `derived_user_properties` means we overwrite the old values for old sessions, which is not right for our data. For example, our shipping time estimates are per-session.

So I'd like to add `derived_session_properties`.

Do you think it's worth waiting for #46 to go in before working on this? That looks pretty major so some risk of a merge conflict, but also looks a bit stalled?

willbryant:
Fixes #54.

## Description & motivation

`derived_user_properties` is excellent and exactly what I needed for some of my properties that once set don't really change.

But we also have some event params that will most likely be different for each session, and putting them in `derived_user_properties` means we overwrite the old values for old sessions, which is not right for our data. For example, our shipping time estimates are per-session.

This PR implements a comparable `derived_session_properties`.

I debated whether these should go in the session dimension or fact table, but I see that the TODO says we should merge them anyway so I didn't stress about it, and to me this is more likely to be facts than dimensions.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Features include:
 | stg_ga4__event_to_query_string_params | Mapping between each event and any query parameters & values that were contained in the event's `page_location` field |
 | stg_ga4__user_properties | Finds the most recent occurance of specified user_properties for each user|
 | stg_ga4__derived_user_properties | Finds the most recent occurance of specific event_params and assigns them to a user's user_key. Derived user properties are specified as variables (see documentation below) |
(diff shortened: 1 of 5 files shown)


adamribaudo-velir (approved):

adamribaudo-velir:
Great idea and execution. And thanks for adding a unit test!

PR #59, opened 2022-09-04
Base model partition optimization fixes

adamribaudo-velir:
## Description & motivation
Related to #40 , updated `stg_ga4__events` to remove the row number window function that was causing all base partitions to be scanned on every query. 

Also removed a window function from `stg_ga4__event_page_view` which had a similar effect on that model. 

Some discussion in #40 on how we should handle duplicate rows moving forward. The `event_key` field is now a surrogate key based on `session_key`, `event_name`, and `event_timestamp`, but that doesn't guarantee uniqueness. 

## Checklist
- [x ] I have updated the README.md (if applicable)

### DIFF
--- a/models/marts/core/fct_ga4__pages.sql
+++ b/models/marts/core/fct_ga4__pages.sql
@@ -9,7 +9,6 @@ with page_view as (
         count(distinct user_key ) as users,
         sum( if(ga_session_number = 1,1,0)) as new_users,
         sum(entrances) as entrances,
-        sum(exits) as exits,
         sum(engagement_time_msec) as total_time_on_page 
 from {{ref('stg_ga4__event_page_view')}}
     group by 1,2,3,4,5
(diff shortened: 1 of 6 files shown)


adamribaudo-velir:
@dgitis this doesn't solve the point your raised about the cost to run jobs, but it does at least provide the opportunity to limit scanning of partitions in downstream models so I think it's a step in the right direction. Let me know what you think.

models/marts/core/fct_ga4__pages.sql
```diff
@@ -9,7 +9,6 @@ with page_view as (
         count(distinct user_key ) as users,
         sum( if(ga_session_number = 1,1,0)) as new_users,
         sum(entrances) as entrances,
-        sum(exits) as exits,
```
dgitis, line 12: Why drop this? I don't personally love it as a metric, but people tend to ask for whatever they see is missing from other reports and I think it's easier to have the metric than to explain why there are better things to look at.

dgitis (approved):
This is a good start. Exits probably isn't worth the costs. Previous page will be missed, but the best place to implement it is in GTM.

PR #64, opened 2022-09-07
Incremental session fact table that includes conversion metrics

adamribaudo-velir:
## Description & motivation

Previously, the `fct_ga4__sessions.sql` model scanned all underlying event partitions. Now it, there is a `fct_ga4__sessions_daily` partitioned table that is much more efficient. The `fct_ga4__sessions` model aggregates across days and is less efficient, but more practical. 

Derived session properties were moved from `fct_ga4_sessions` to `dim_ga4__sessions`.

## Checklist


adamribaudo-velir:
TODO: look into whether the session mart model(s) can be partitioned on session start date, rather than using session_key as the merge key.

willbryant:
Just saw this one, love where this is going!

dgitis:
The branch is way too far off of Main to ever merge back in, but I
partitioned on session_start_date in the first-session branch and have that
deployed to a high-volume client right now. It works great. We should
definitely partition this table.


*Damon Gudaitis*
www.caretjuice.com

adamribaudo-velir:
After reworking this, I think I lost track of the fact that the flaw in the `insert_overwrite` incremental strategy is that we could accidentally pick up the tail end of a session that lasts >24hrs and overwrite the session metrics. Ex: A session lasts from 5PM to 12:01AM and we only pick up the metrics from 12:00AM-12:01AM and overwrite the session_key with those metrics. 

I had previously tried to use the `merge` incremental strategy with `session_key` as the merge key, but there was no way to limit the underlying partition scans on `event_date_dt` in that scenario. From the DBT docs https://docs.getdbt.com/reference/resource-configs/bigquery-configs#the-merge-strategy

"The merge approach has the benefit of automatically updating any late-arriving facts in the destination incremental table. The drawback of this approach is that BigQuery must scan all source tables referenced in the model SQL, as well as the entirety of the destination table. "

So stuck between a rock and a hard place. Open to suggestions.

adamribaudo-velir:
One approach would be to recognize that this a 'fact' table. We don't need each row to be unique on `session_key` as long as the end consumer of this model can sum across rows and then join in the 'dimension' table. This frees up a lot of pressure because then it's not problematic to have 2 rows for 1 session covering 2 different days. 

Ex: For a session that lasts from 5PM-12:01AM, row 1 counts the metrics from 5PM-12AM, row 2 counts the metrics from 12AM-12:01AM. If you aggregate on session_key, you'll get the correct totals.

willbryant:
Yeah seems like a reasonable direction. What about having a single fct_ga4__daily_sessions table (replacing dim_ga4__sessions and fct_ga4__sessions) plus a fct_ga4__sessions view that combines those daily sessions back up to the overall level?

Then we can do the heavy lifting for users in terms of figuring out what should be min, what should be sum, what should be first_value, etc.

adamribaudo-velir:
Implemented your suggestion of a `fct_ga4__sessions_daily` table, @willbryant . Thanks for the idea. I think that gives the option for performance that some may need, while the `fct_ga4__sessions` table is less efficient, but practical for other cases. 

Going to merge this and create a new release.

Issue #68, opened 2022-09-14
Google Paid Misattributed as Organic


dgitis:
We've discussed this possibility before. Because Google Ads traffic is attributed through a server-side integration, the BigQuery export doesn't properly assign source, medium, campaign and default_channel_grouping attributing them to organic sources instead.

This can be verified using the following query.

```
SELECT  
  session_key,
  ga_session_number,
  landing_page,
  source,
  medium,
  campaign,
  default_channel_grouping
FROM `*.dim_ga4__sessions` 
WHERE source = 'google'
and landing_page like '%gclid%'
```

PR #70, opened 2022-09-15
Fix Google Ads Attributed to Organic

dgitis:
## Description & motivation
Because Google Ads passes attribution data through a server-to-server integration with Google Analytics, Google Ads traffic was being misattributed as `google / organic` in source, medium, and default_channel_grouping fields as identified in issue #68. 

This PR is a preliminary fix to this problem. It detects gclid in the query string in stg_ga4__events and rewrites the medium and campaign as 'cpc' and '(cpc)' respectively. This improves the accuracy of default_channel_grouping field which is calculated from the fields in stg_ga4__events.

This is not, however, a complete fix. Here are the known issues and my reasons for not fixing them.

- Paid channels may not be accurately assigned. Google lists five different channels for Google Ads traffic: Paid Search, Paid Video, Display, Cross-network, and Paid Social. The rules for assigning these channels come from the Google Ads-Analytics integration so we can't copy their logic. Instead we have to use URL parameters and page referrers to derive our rules. However, Google stores the same information in different parameters depending on circumstances that require extensive reverse engineering. Because Google obviously changes these URL parameters, any work would be fragile requiring updates every time Google changes its URL parameter schema.
- Google display traffic is should be 'display' instead of 'cpc.' For the same reasons as above, I left this until someone has the time (and access to multiple Google Ads accounts linked to BigQuery via GA) to explore this more in-depth.
- The Google Ads campaign sometimes appears in query parameters. However, there are at least two query parameters that can have the campaign in them and not all query strings have the campaign in them. As with everything else here, this requires extensive investigation.

## Checklist
- [ n/a ] I have updated the README.md (if applicable)


seeds/ga4_source_categories.csv
```diff
@@ -817,4 +817,188 @@ zalo,SOURCE_CATEGORY_SOCIAL
 zen.yandex.ru,SOURCE_CATEGORY_SEARCH
 zoo.gr,SOURCE_CATEGORY_SOCIAL
 zooppa,SOURCE_CATEGORY_SOCIAL
-zooppa.com,SOURCE_CATEGORY_SOCIAL
\ No newline at end of file
+<<<<<<< HEAD
```
adamribaudo-velir, line 820: I think you may have committed some merge conflict text here?

models/staging/ga4/stg_ga4__events.sql
```diff
@@ -61,7 +61,18 @@ enrich_params as (
         {{extract_hostname_from_url('page_location')}} as page_hostname,
         {{extract_query_string_from_url('page_location')}} as page_query_string,
     from remove_query_params
+),
+fix_params as (
+    select
+        * except (medium, campaign),
+        case
+            when page_query_string like '%gclid%' then "cpc"
```
adamribaudo-velir, line 69: I think I'm missing what the difference is between these 2 case statements. Aren't they both searching for the same thing? `page_query_string like '%gclid%'

adamribaudo-velir:
I think I'm on board, I'm just seeing some weird behavior now and have to move on to something else:

- I'm running the code in this PR, but am still getting google / organic for this page_location: https://www.velir.com/?sc_camp=7C14B25B156A4CF29D33D031A291A561&gclid=Cj0KCQiApL2QBhC8ARIsAGMm-KFC_aQ2byVY6-
- I have `gclid` set in my `query_parameter_exclusions` project setting, so I shouldn't be seeing `gclid` in the page_location at all
- We're running the `detect gclid` CTE after the query parameters are excluded. We should run it before (or run it off the original page_location

adamribaudo-velir (approved):

PR #71, opened 2022-09-28
Session event count metrics

dgitis:
## Description & motivation
This allows you to add event counts to the `fct_ga4__sessions` table in addition to the default count_page_views.

You configure it with an array custom parameter that matches event names and performs a `countif(event_name = 'array_item') as count_array_item`

## Checklist
- [ n] I have added tests & descriptions to my models (and macros if applicable)
- [ n] I have run `dbt test` and `python -m pytest .` to validate exists tests


dgitis:
I realize that session event counts is the same thing as session conversions.

PR #73, opened 2022-10-19
Fix dim_ga4__sessions unique test

adamribaudo-velir:
Fixes the failing unique session_key test on dim_sessions by filtering on row_number = 1 to account for duplicate session_start events. 

We had previously counted row numbers in stg_ga4__events but that was removed to help with partition scanning. The downstream effect of that was allowing duplicate session_start events which caused duplicates in `dim_ga4__sessions`

Long-term, we may want a more generic model that counts event row numbers if this is useful for other models. Also, I would prefer not to use `session_start` as the basis for `dim_ga4__sessions` given how unreliable it is as an event, but there arenn't many compelling alternatives.

## Checklist


dgitis (approved):
We should also consider for future iterations that there are sessions without a session_start event. This is still an improvement.

Issue #69, opened 2022-09-14
dim_ga4__sessions model relies on flaky session_start event


dgitis:
It should use the sessions_first_last_pageview model instead.

adamribaudo-velir:
@willbryant you seem to be using the session marts. Do you have an opinion here? Is the `session_start` event a useful way to define sessions if it isn't fired consistently by all browsers? Or would you rather use `session_id` + the first `page_view` event?

willbryant:
@adamribaudo-velir I personally think that finding sessions by grouping up the session keys sounds simpler and sounder. The docs do say that the `session_start` event will sometimes get lost.

However, I have a thought regarding the suggestion to use the pageview model: this won't work for mobile app (Firebase) event streams.

So, I guess my thinking is that it would need to be based on simply all events, unless we're going to fish out pageviews + some representative mobile event stream.

Issue #66, opened 2022-09-09
Uniqueness test failures on main


willbryant:
Since updating to fdd19c2f13ad4fcf3b400a237a813a417617a195, the following tests are failing:

```
10:45:43  Failure in test unique_stg_ga4__users_first_last_events_user_key (models/staging/ga4/stg_ga4.yml)
10:45:43    Got 452 results, configured to fail if != 0
10:45:43  
10:45:43    compiled SQL at target/compiled/ga4/models/staging/ga4/stg_ga4.yml/unique_stg_ga4__users_first_last_events_user_key.sql
10:45:43  
10:45:43  Failure in test unique_dim_ga4__sessions_session_key (models/marts/core/core.yml)
10:45:43    Got 11 results, configured to fail if != 0
10:45:43  
10:45:43    compiled SQL at target/compiled/ga4/models/marts/core/core.yml/unique_dim_ga4__sessions_session_key.sql
10:45:43  
10:45:43  Failure in test unique_stg_ga4__users_first_last_pageviews_user_key (models/staging/ga4/stg_ga4.yml)
10:45:43    Got 15 results, configured to fail if != 0
```

willbryant:
@adamribaudo-velir Keen to hear your thoughts on this.

In that commit you removed the uniqueness test on the events table. But these other tests still exist.

I'm not sure if this is the right direction to be going anyway?

adamribaudo-velir:
Right, a few things are held up and waiting on a partition-friendly way to keep events unique. Using `row_number() over(partition by session_key)` negates the partition pruning and causes unnecessary scans for any model accessing `stg_ga4__events.sql`

If we can find a way to determine event uniqueness, this feels like the right direction to take. But leaving `main` in a state with failed tests isn't great, so apologies for that. 

Any ideas on event uniqueness?

willbryant:
Is the original problem that there are duplicate deliveries of the same event, or that there is sometimes more than one session start event per session?

willbryant:
Aargh, I'm running into #61 again, so my sessions are just not working properly at all.

Specifically, I am missing rows from the dim_ga4__sessions table corresponding to the fct_ga4__sessions and other events, because the session_start event is associated with a session_key that is based on a user_key that is sometimes based on user_pseudo_id and sometimes based on user_id (from later in the session once they've logged in). And the dim_ga4__sessions table is based on the session start events, therefore, it's affected.

[...]

PR #74, opened 2022-10-19
Fix to make event_key unique

adamribaudo-velir:
## Description & motivation

Tests were failing due to duplicate `event_key` values. This has been resolved by including `event_params` in the surrogate key. 

Rationale for not using window functions: This image demonstrates how including a row_number() window function causes BQ to bill for all records in the underlying base table (68MB of records in this case)

image

## Checklist

### DIFF
--- a/integration_tests/test_stg_ga4__events.todo
+++ b/integration_tests/test_stg_ga4__events.todo
@@ -1,3 +1,5 @@
+# Test test currently fails because the event_key depends on the event_params nested field. Cannot mock that using CSV. 
+
 import pytest
 from dbt.tests.util import read_file,check_relations_equal,run_dbt
 
@@ -37,5 +39,5 @@ class TestStgEvents():
     
     def test_mock_run_and_check(self, project):
         run_dbt(["build"])
-        #breakpoint()
+        breakpoint()
         check_relations_equal(project.adapter, ["actual", "expected"])
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -18,17 +18,18 @@ add_user_key as (
         end as user_key
     from base_events
 ), 
--- Add unique keys for sessions and events
+-- Add unique key for sessions. session_key will be null if user_pseudo_id is null due to consent being denied. ga_session_id may be null during audience trigger events. 
 include_session_key as (
     select 
         *,
-        to_base64(md5(CONCAT(stream_id, CAST(user_key as STRING), IFNULL(CAST(ga_session_id as STRING), 'no_session')))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
+        to_base64(md5(CONCAT(stream_id, user_pseudo_id, IFNULL(CAST(ga_session_id as STRING), 'no_session')))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
     from add_user_key
 ),
+-- Add unique key for events
 include_event_key as (
     select 
         include_session_key.*,
-        to_base64(md5(CONCAT(CAST(user_key as STRING), session_key, event_name, CAST(event_timestamp as STRING), to_json_string(event_params)))) as event_key -- Surrogate key for unique events. These keys may not be unique given how GA4 operates. 
+        to_base64(md5(CONCAT(session_key, event_name, CAST(event_timestamp as STRING), to_json_string(event_params)))) as event_key -- Surrogate key for unique events. These keys may not be unique given how GA4 operates. 
     from include_session_key
 ),
 detect_gclid as (
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -25,13 +25,6 @@ include_session_key as (
         to_base64(md5(CONCAT(stream_id, CAST(user_key as STRING), cast(ga_session_id as STRING)))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
     from add_user_key
 ),
-/*
--- Removing row_number() calculation as it negates the base table partition. Would like to include equivalent method of determining event uniqueness somehow
-include_event_number as (
-    select include_session_key.*,
-        row_number() over(partition by session_key) as session_event_number -- Number each event within a session to help generate a uniqu event key
-    from include_session_key
-),*/
 include_event_key as (
     select 
         include_session_key.*,
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -28,7 +28,7 @@ include_session_key as (
 include_event_key as (
     select 
         include_session_key.*,
-        to_base64(md5(CONCAT(CAST(session_key as STRING), event_name, CAST(event_timestamp as STRING), CAST(event_previous_timestamp as STRING)))) as event_key -- Surrogate key for unique events
+        to_base64(md5(CONCAT(CAST(user_key as STRING), CAST(ISNULL(session_key, 0, session_key) as STRING), event_name, CAST(event_timestamp as STRING)))) as event_key -- Surrogate key for unique events
     from include_session_key
 ),
 detect_gclid as (
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -13,8 +13,8 @@ add_user_key as (
         *,
         case
             when user_id is not null then to_base64(md5(user_id))
-            when user_pseudo_id is not null then to_base64(md5(user_pseudo_id))
-            else null -- this case is reached when privacy settings are enabled
+            when user_pseudo_id is not null and user_pseudo_id != '' then to_base64(md5(user_pseudo_id))
+            else null -- this case is reached when privacy settings are enabled and (possibly) for certain audience triggered events
         end as user_key
     from base_events
 ), 
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -22,13 +22,13 @@ add_user_key as (
 include_session_key as (
     select 
         *,
-        to_base64(md5(CONCAT(stream_id, CAST(user_key as STRING), cast(ga_session_id as STRING)))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
+        to_base64(md5(CONCAT(stream_id, CAST(user_key as STRING), IFNULL(CAST(ga_session_id as STRING), 'no_session')))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
     from add_user_key
 ),
 include_event_key as (
     select 
         include_session_key.*,
-        to_base64(md5(CONCAT(CAST(user_key as STRING), IFNULL(session_key, 'no_session'), event_name, CAST(event_timestamp as STRING)))) as event_key -- Surrogate key for unique events
+        to_base64(md5(CONCAT(CAST(user_key as STRING), session_key, event_name, CAST(event_timestamp as STRING)))) as event_key -- Surrogate key for unique events
     from include_session_key
 ),
 detect_gclid as (
--- a/models/staging/ga4/stg_ga4__events.sql
+++ b/models/staging/ga4/stg_ga4__events.sql
@@ -28,7 +28,7 @@ include_session_key as (
 include_event_key as (
     select 
         include_session_key.*,
-        to_base64(md5(CONCAT(CAST(user_key as STRING), session_key, event_name, CAST(event_timestamp as STRING)))) as event_key -- Surrogate key for unique events. These keys may not be unique given how GA4 operates. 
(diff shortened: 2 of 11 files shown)


dgitis (changes requested):
I can't find a single GA4 property where event_previous_timestamp has any other value than null.

Please test these changes on a large site to see if they do anything.

adamribaudo-velir:
Good catch, I didn't think that through. 

I finally have some clients on GA4 that I can test with. Looking at a relatively large site, there are plenty of examples of duplicate events. 

One example: I have 13 click events with identical timestamp, event name, etc. Only thing that's different is `engagement_time_msec`. But even if I add that into the surrogate key, I get back 14k events that fail the uniqueness test. 

So I'm at a loss. We could run a distinct on all non-nested fields to dedupe events as a service to the end-user. I have a slight concern that we could lose distinct information contained in the nested fields, though. 

Suggestions?

adamribaudo-velir:
@willbryant @dgitis I've dug into this quite a bit and have the following changes proposed in this PR. Description has been updated to reflect my thinking. Any concerns?

This PR fixes all defined unique tests, btw.

dgitis:
We discussed this months ago and rejected it, but maybe we look again at
deduplicating events with the same timestamp or even just page views. We'd
lose a tiny fraction of event tracking accuracy from legitimate duplicates,
gain accuracy from accidental duplicates, and solve a bunch of problems.


*Damon Gudaitis*
www.caretjuice.com

adamribaudo-velir:
@dgitis The only way I can think to remove duplicates would be to group by all non-nested fields (everything except `event_params` and `user_properties`. But then you lose the 2 nest fields. 

Can you think of another way to remove duplicates that doesn't involve window functions?

adamribaudo-velir:
A suggestion came in from #measure Slack around including `to_json_string(event_params)` in the surrogate key and it worked for the 2 websites that I tested. Can anyone else test with their data?

Run the project and then test with `dbt test -m stg_ga4__events`

models/staging/ga4/stg_ga4__events.sql
```diff
@@ -13,29 +13,22 @@ add_user_key as (
...
+            else null -- this case is reached when privacy settings are enabled and (possibly) for certain audience triggered events
         end as user_key
     from base_events
 ), 
 -- Add unique keys for sessions and events
 include_session_key as (
     select 
         *,
-        to_base64(md5(CONCAT(stream_id, user_pseudo_id, ga_session_id))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
+        to_base64(md5(CONCAT(stream_id, CAST(user_key as STRING), IFNULL(CAST(ga_session_id as STRING), 'no_session')))) as session_key -- Surrogate key to determine unique session across streams and users. Sessions do NOT reset after midnight in GA4
```
willbryant, line 25: Is the revert back to using user_key instead of user_pseudo_id deliberate here?

I don't think the CASTs are necessary BTW, aren't all those things strings? I don't remember ever seeing a NULL ga_session_id either personally.

willbryant:
Would it be possible to do the de-duping on the way into the base events table, rather than the staging table reads?

This would have the advantage that, thanks to the incremental build process, we are only ever dealing with one source date table at once, so the partition pruning issue becomes moot.

models/staging/ga4/stg_ga4__events.sql
adamribaudo-velir: Right, use of user_key was an oversight. I'll update that. 

`ga_session_id` can definitely be `null`. See audience triggers: https://support.google.com/analytics/answer/9934109?hl=en

adamribaudo-velir:
@willbryant if you have a suggestion on how to handle de-duping on the way to the base events table, I'd love to hear it. I've given it some thought and have come up short. Were you able to confirm whether the event_keys are now unique? 

The use case for partition pruning is being able to create incremental tables on top of `stg_ga4__events`. For example, in the future we'd like the session mart to be incremental which means it will have to filter on `event_date_dt` to find the latest `session_start` events. Just wanted to make sure that was clear.

willbryant:
I think my datasets must be much smaller than you guys. If I include the event params in the grouping criteria, then I don't see any duplicate events from my web stream, at least in the few source table dates I've spot-checked. I do see them very regularly from Firebase events though, which is interesting.

willbryant:
@adamribaudo-velir I think I probably just don't understand the issue, I haven't tried it. Is there something stopping us from doing a group by/distinct on the load from the raw events tables to the base table?

I thought it was the event params/user properties type arrays, but you've sort of solved that with the to_json_stringify solution.

What else have I missed 😂

adamribaudo-velir:
I think there's nothing left to understand because it's solved now that it's clear that `event_params` can be used to dedupe :) The only reason this is occurring in the `stg` model and not `base` is due to DBT conventions stating that `base` models shouldn't modify the raw data aside from renaming and light conversions.

willbryant:
Ah, I see. I thought the current guidance was that *sources* shouldn't be messing with the data, not base tables.

In any case, even if it needs a rename and/or an extra layer, would materializing and de-duping "on the way in" instead of "on the way out" help us?

(But, raised #80 with an existing issue that might be relevant in the decision.)

dgitis:
I ran this revision against 43 GB of data from a base partition. Unfortunately, this revision didn't pass.

image

I'll try to track down what events didn't pass tomorrow.

dgitis:
For reference, that site has an ad_impression event that fires a lot. If it's the lone cause then I'd approve this PR as being better than the status-quo. Failing 202 times in this volume of data is pretty darn close to zero.

willbryant:
The other big array-of-struct field is user_properties, so it would be interesting to see if that field varies for those near-duplicates.

dgitis:
I'm thinking that if we want to consider deduplicating events and making
events unique at the same time, we might want to drop events with duplicate
event_params arrays. I'm fine work possible edge cases like mine (pending
confirmation) requiring extra implementation effort.

adamribaudo-velir:
There's a discussion about this in the #measure Slack instance and this was very interesting:

"There’s nothing in those individual events that determine a sequence number or a timestamp – there’s no deterministic order either. Essentially, hits that happen within a batch are completely devoid of any information as to when those events actually happened AND what their relative order is. That’s why sending a timestamp as a custom event parameter with all hits is very useful if you want to"

So if you want distinct events, you need to update your GA4 implementation to include a sequence or random int or original timestamp. It can't be solved entirely in BQ.

dgitis:
@willbryant's idea to perform this between the source data and the base table has a lot of merit.

When we deliberately set the partitions that we are querying, window functions only operate on those partitions. Querying, yesterday's data with `where event_date_dt in (current_date, date_sub(current_date interval 1 day))` will force the window function to only scan yesterday's and today's partitions.

This could solve a lot of problems that we are having. We could move all window functions here, like the session_first_last models and add session_first_page_view, session_last_page_view etc... to each event, do the row_number calculation and not cause failures on the unique test for event_key. 

We could fix session_start in some way by generating a session_start event where it is missing or selecting the earliest event with referrer info populated and use that for first_event calculations possibly adding something like session_first_event_key.

Basically, we process all of our window functions between the two partitioned tables and the `stg_ga4__events` table retains its fluid sizing which is convenient for ad-hoc queries against staging tables.

dgitis:
Maybe we could add a client_side_timestamp_field variable. Users working on large sites, can use this variable to map any custom timestamp, as discussed in #measure Slack, and, if this variable is set, we add it to the event_key. 

I like this solution a bit better than the status quo.

willbryant:
Apologies if this is an obvious point but re uniqueness, I only personally had an issue when the defined test started failing. That is, no-one was complaining about non-unique events from GA4, which is never 100% perfect data anyway; just that the DBT pipeline failed.

What I'm getting at here is that if we generate event_key I would expect it to be unique - ie. I would expect the de-duping to match up with the event_key generation, whatever that is.

I don't personally have any strong opinion on what should go into event_key, as clearly I am a smaller user than you guys and don't run into the same level of issues.

willbryant:
Re session_start, I'm definitely keen on generating this based on first events rather than session_start events. (As I commented in #69, I think this needs to be from _any_ event, not just page-view events, in order to cover mobile/Firebase.)

I like your direction in https://github.com/Velir/dbt-ga4/pull/74#issuecomment-1290997489. My comments around the intra-day headaches in #80 may not be a blocker - Google explicitly documents the intraday data as incomplete and non-final in a number of respects, so I think people will be OK with it being not cleaned up the same way.

Support for an optional client-side de-duping parameter makes sense to me. One thought, perhaps rather than hard-coding this sort of thing, we should be allowing people to configure the variables that go into event_key, and then we can just document how you would go about utilizing an extra custom param to do extra uniqueness control. This may give others the flexibility solve different issues too.

dgitis:
The dbt build command that is used by default in dbt cloud runs the tests just after each model is built while other threads build and test other models. When a test fails, dbt skips any child models that rely on the model that failed the test. Because the stg_ga4__events table is a requirement for every other table, we need our solution, whatever it is, to actually work.

I'm also okay with removing the test and settling with what we have now.

dgitis:
For session_start, if we generate a session_start event, we will also need to deduplicate as missing and duplicate session_start events are both problems. If we go with replacing session_start entirely, we might want to consider using page_view or screen_view (and make this our default everywhere we have page_view to support mobile) and possibly add a variable to add a list of valid session events to support measurement protocol "sessions".

I had intended the de-duping parameter to be configurable.

willbryant:
What I had in mind was not just configuring that one parameter, but configuring the entire list, eg.

```
{{ dbt_utils.surrogate_key(var('event_key_parameters', ['session_key', 'event_name', 'event_timestamp', etc])) }} AS event_key
```

(with whatever the current thinking is on the default list of event_key fields should be)

I'm using surrogate_key here which seems to be the standard way of doing hashing now, but mainly because it takes care of array arguments for us.

dgitis:
That looks good to me. When we document it, either in readme.md or in a comment, we could recommend adding a timestamp in the client to ensure uniqueness as our intended use of this variable.

adamribaudo-velir:
- If the goal is to provide unique event_keys for the majority of users who have a small-medium implementation, this PR does that. For the remaining users with large implementations, they can update their GTM implementation to include a unique value in the event params and this PR will automatically incorporate that into the event_key without the need for an additional `client_side_timestamp_field` variable. So I'm not following why the surrogate key for events would need to be configurable. Why would anyone need fewer fields in the surrogate key? Why would anyone need more event params in the surrogate key if we already include all event params?

- If there's a suggestion to 'dedupe on the way in' I'm definitely open to it, I just need to see some code to fully understand the implementation. Can someone open a PR (even a draft one) with those changes? I'm just not following what that would look like. 

This discussion is getting pretty long, I'm hoping that someone can provide an alternative PR so we can have a more informed discussion about the merits.

willbryant:
Ah yeah duh, good point re event params.

Testing out this branch at b3dd3d277f0ba6df9e1b9f5f6ee531e81743e8e4 against just my small non-prod dataset first, I get:

```
20:25:44  Failure in test unique_dim_ga4__sessions_session_key (models/marts/core/core.yml)
20:25:44    Got 19 results, configured to fail if != 0
20:25:44  
20:25:44    compiled SQL at target/compiled/ga4/models/marts/core/core.yml/unique_dim_ga4__sessions_session_key.sql
```

integration_tests/test_stg_ga4__events.todo
```diff
@@ -37,5 +39,5 @@ def models(self):
     
     def test_mock_run_and_check(self, project):
         run_dbt(["build"])
-        #breakpoint()
+        breakpoint()
```
willbryant, line 42: Revert?

willbryant:
With that commit the session test passes, and all tests pass on my non-prod dataset. On my prod dataset (~104 GB for base_ga4__events), I get:

```
21:01:17  Failure in test unique_stg_ga4__events_event_key (models/staging/ga4/stg_ga4.yml)
21:01:17    Got 58258 results, configured to fail if != 0
21:01:17  
21:01:17    compiled SQL at target/compiled/ga4/models/staging/ga4/stg_ga4.yml/unique_stg_ga4__events_event_key.sql
```

willbryant:
I pulled the full rows for an event_key that had 3 rows, and the json dump was exactly the same for all 3. So they were true duplicates. I then had a look at the event names of all the non-unique rows:

Screen Shot 2022-10-27 at 10 11 25

Looks like those are all firebase events.

adamribaudo-velir:
Thanks for checking! I updated the readme with our notes.

integration_tests/test_stg_ga4__events.todo
adamribaudo-velir: I dropped this pytest for a few reasons I can get into. It was renamed to *.todo to keep the code handy but it won't run as part of the test suite. Just leaving it for future refactoring if I/we decide it adds value.

willbryant:
Unfortunately in this case the events are generated by Firebase itself, so injecting those parameters isn't available as a solution.

I have non-unique events with the current main branch, so if you wanna go ahead and merge this PR in the meantime, no reason to let my feedback delay you. But personally the #66 issue is unfortunately still not resolved (and perhaps needs to remain so until we can try another de-duping approach).

adamribaudo-velir:
shoot, I see what you mean. That reminds me that there are some automatic gtag.js events that could produce the same situation as well. 

Ok yea, I'd like to merge this PR but leave #66 open and maybe summarize what we've learned so far in that issue.

willbryant (approved):

Issue #62, opened 2022-09-05
Create user_pseudo_id to user_id mapping table. Replaces current user_key implementation. 


adamribaudo-velir:
As @willbryant found, user_id's can change mid sessions which then cause user_keys to change mid-session. The intention of the `user_key` is to have a stable identifier for users as they they move from 1 to more-than-1 device. The current implementation is too brittle.

A better implementation would use a mapping table to keep track of user_pseudo_ids and user_ids. The mapping table would then be used in the marts when users need to be related to sessions. user_pseudo_id will be a sufficient field to use for session and page-level metrics. Only `user` level metrics will need this mapping table. 

My suggestion would be to only support mapping between a `pseudo_user_id` to 0 or 1 `user_id`. In this implementation, we would track the 'last seen' user_id for each pseudo_user_id, rather than attempting to merge multiple user_ids into a single entity.

PR #76, opened 2022-10-20
Use user_pseudo_id rather than user_key to construct session_key (#61)

### LINKED ISSUE #61: Events not associated with session correctly when user_id changes
We're noticing that some events are associated with `session_key` values that do not appear in `dim_ga4__sessions`.

If I run this query:

```
SELECT event_date_dt, event_name, user_pseudo_id, user_id, user_key, ga_session_id, session_key
from `delivereasy-data-prod`.`bi_dbt_ga4`.`stg_ga4__events`
where user_pseudo_id = '1hzXd+Z2R1....' and event_date_dt = '2022-09-01'
order by event_timestamp
```

We can see why:

Screen Shot 2022-09-05 at 22 49 58

GA4 has not managed to copy the `user_id` back to the start of the session (which I thought it was meant to do TBH, but there you go).

Therefore, the `user_key` has been reset when the user logs in and their `user_id` goes from `NULL` to set. This, in turn, causes `session_key` to change.

Finally, because this new session does not have a `session_start` event, because it is not a new session from GA4's perspective, the second session is generated in the `fct_ga4__sessions` table but *not* in the `dim_ga4__sessions` table.

As a result, queries that join across to `dim_ga4__sessions` drop the event rows (or miss data, if we use a left join).

willbryant:
Fixes #61.

## Description & motivation

user_key can vary across a session, because user_id can vary across a session. Therefore, using user_key to build session_key broke sessions in two whenever user_id changed (eg. when a user logged in).

This commit fixes this by:
1. using user_pseudo_id instead of user_key to build the session_key
2. using window functions to find the last user_key for the sessions facts table, instead of grouping (as there can now be multiple user_keys in a single session's events).

This should keep us going until #62 is done.

## Checklist


adamribaudo-velir (approved):
I like it. We could conceivably create a mart for 'devices' using pseudo_user_id and then a mart for 'users' using a mapping of all known pseudo_user_ids that map to user_ids.

PR #82, opened 2022-10-27
Simplify derived user properties & session properties queries

willbryant:
## Description & motivation

Use LAST_VALUE with IGNORE NULLS and DISTINCT instead of a set of extra CTEs and joins.

To avoid 'Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex.' errors.

## Checklist


adamribaudo-velir (approved):
Nice work. And cool to see the related unit tests adding value by demonstrating that nothing broke.

Issue #80, opened 2022-10-25
Incremental + streaming strategy behaves oddly when tables are rotated


willbryant:
@adamribaudo-velir @dgitis since we are discussing the event loading model, I will open this issue for another headache that my business users noticed last week.

At somewhere around ~10-11am my local time, GA4 creates the new `events_` table for the previous day, and some time after that drops the `events_intraday_` table for the same date.

What they noticed is that the events for today *are* present throughout the day, and up to that magic time the next morning, but then disappear!

The cause is of course that I haven't re-run the DBT job, and whereas the streaming/intraday table is a direct view onto `events_intraday_*`, the non-streaming base table isn't a view onto `events_*`, the events have actually been copied out.

I don't have a suggestion on how to solve this yet. The best I could do right now is set up the glue necessary to rerun the DBT pipeline when the new `events_*` table is created, making sure we are de-duping with the intraday table successfully (which may be the other fatal flaw in my idea of de-duping on the way into the base events table).

The other approach is to give up on direct access to the intraday tables, and expect a DBT run to pull in that data too. Obviously that is in many respects a step backwards, but to be honest, it's what I actually expected before using this project - most of my downstream models have to be materialized, so I had assumed I would only be getting updates when I ran the pipeline.

willbryant:
I've been struggling with this one for a couple of days and the best I can come up with is to replace the current `events_intraday_*` wildcard with an `events_*` wildcard, and use `_TABLE_SUFFIX` to avoid pulling in all the dates already partitioned.

eg. let's say I ran the pipeline last night, 2022-10-26. The last non-intraday table at that point was the previous day's `events_20221025`. I now want a query that will pick up 2022-10-26 (and 2022-10-27 when it comes), whether the 26th is in `events_intraday_20221026` or `events_20221026`.

So I do this by querying `events_*` with a `WHERE _TABLE_SUFFIX > '20221025'` condition.

(I don't need to use `WHERE _TABLE_SUFFIX > '20221025' OR _TABLE_SUFFIX > 'intraday_20221025'` because the `i` of intraday is `>` the `2` of `20221025`, so the one condition works to find both.)

[...]

Issue #33, opened 2022-07-19
Keeping base_ga4__events_intraday and base_ga4__events in sync is not DRY


adamribaudo-velir:
We have duplicated select statements in `base_ga4__events_intraday` and `base_ga4__events`. This is because they need to be unioned together and `base_ga4__events_intraday` needs to remain a view while `base_ga4__events` needs to remain a table. 

This has caused a few issues at this point when the statements aren't kept in sync. Generally speaking, we want to keep code DRY (don't repeat yourself) which can avoid issues. 

Any ideas for how to keep them aligned? 

One idea: Use a macro that spits out the select statement. No input parameters necessary, just have it output text.

dgitis:
A macro seems like the only way. I'm also thinking we might want a config
macro which would go well with merging the two base_ga4 models.

[image: Damon Gudaitis, Principal at Caret Juice Marketing]

*Damon Gudaitis*
Principal, Caret Juice Marketing
www.caretjuice.com
e: ***@***.*** | p: 778-846-9248

dgitis:
I have a fix for this issue planned in #107 which is being worked on in the multi-site branch.

PR #83, opened 2022-10-30
De-dup events in the base events queries (#74)

willbryant:
## Description & motivation

Fix duplicate event errors as discussed with @adamribaudo-velir and @dgitis in #74.

Previously we did this in the stg_ga4__events view, but that turned out to be incompatible with parition pruning, an important cost optimization.

Doing it in the "base" queries does not suffer from this problem because the incremental update process ensures that we are only querying one date partition at a time anyway, which is the right granularity to run de-duping at because the duplicate events have duplicate event_date_dt/event_timestamp values and therefore get delivered into the same partition.

## Checklist

Benchmarks below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,10 +227,6 @@ vars:
     conversion_events:['purchase','download']
 ```
 
-# Event Key Uniqueness
-
-The `stg_ga4__events` model generates a surrogate key for each event using a combination of the session key, event name, event timestamp, and the event parameters. Given how GA4 (and specifically gtag.js) work, however, it is not uncommon to see multiple events containing exactly the same timestamps and parameters. This in turn creates duplicate `event_key` values and can introduce downstream issues when joining on `event_key`.  Running `dbt test -m stg_ga4__events` will fail if you have duplicate event keys. In these cases, the recommended approach is to update your collection implementation to include a new event parameter that will guarantee uniqueness for each event. For example, adding a `client_event_timestamp` or a `random_int` event parameter to each event should be sufficient to ensure that each `event_key` is unique.
-
 # Incremental Loading of Event Data (and how to handle late-arriving hits)
 
 By default, GA4 exports data into sharded event tables that use the event date as the table suffix in the format of `events_YYYYMMDD` or `events_intraday_YYYYMMDD`. This package incrementally loads data from these tables into `base_ga4__events` which is partitioned on date. There are two incremental loading strategies available:
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Features include:
 
 | model | description |
 |-------|-------------|
-| stg_ga4__events | Contains cleaned event data that is enhanced with useful event and session keys. See note below regarding cases where event keys may not be unique.|
+| stg_ga4__events | Contains cleaned event data that is enhanced with useful event and session keys. |
 | stg_ga4__event_* | 1 model per event (ex: page_view, purchase) which flattens event parameters specific to that event |
 | stg_ga4__event_items | Contains item data associated with e-commerce events (Purchase, add to cart, etc) |
 | stg_ga4__event_to_query_string_params | Mapping between each event and any query parameters & values that were contained in the event's `page_location` field |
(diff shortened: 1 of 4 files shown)


willbryant:
TL;DR it is about 35-50% slower to build the base events, but it respects table pruning correctly so it processes no more data than the current main branch, and it fixes the duplicate events.

*Main branch*

Full run:

```
21:22:37  1 of 35 OK created incremental model test_dbt_ga4.base_ga4__events ............. [CREATE TABLE (116.5m rows, 106.8 GB processed) in 43.66s]
...
21:22:45  Completed with 1 error and 0 warnings:
21:22:45  
21:22:45  Failure in test unique_stg_ga4__events_event_key (models/staging/ga4/stg_ga4.yml)
21:22:45    Got 60038 results, configured to fail if != 0
```

Incremental run:

```
21:25:12  1 of 35 OK created incremental model test_dbt_ga4.base_ga4__events ............. [SCRIPT (3.3 GB processed) in 39.13s]
...
21:25:20  Completed with 1 error and 0 warnings:
21:25:20  
21:25:20  Failure in test unique_stg_ga4__events_event_key (models/staging/ga4/stg_ga4.yml)
21:25:20    Got 60038 results, configured to fail if != 0
```

Result:

```
SELECT count(*) FROM `<project>.test_dbt_ga4.base_ga4__events` 
116549295
```


*This branch*

Full run:

```
21:14:20  1 of 35 OK created incremental model test_dbt_ga4.base_ga4__events ............. [CREATE TABLE (116.5m rows, 106.8 GB processed) in 67.88s]
...
21:15:15  Completed successfully
```

Incremental run:

```
21:16:22  1 of 35 OK created incremental model test_dbt_ga4.base_ga4__events ............. [SCRIPT (3.3 GB processed) in 53.22s]
...
21:17:14  Completed successfully
```

Result:

```
SELECT count(*) FROM `<project>.test_dbt_ga4.base_ga4__events` 
116461189
```

adamribaudo-velir:
This is awesome. I thought an approach like this would suffer from the same issues we see when using window functions in `stg_ga4__events`, but I think because we're incrementally loading in from a sharded table, as opposed to a partitioned table, we don't suffer the same issue with scanning costs. 

I just have 1 question before we merge: do you know how this behaves when `pseudo_user_id` and `ga_session_id` are null? Both are edge cases (pseudo_user_id is null when consent is denied and ga_session_id is null for audience triggers) so I'm fine merging, but wanted to give it some thought.  

My guess is that when either are null, records will be grouped into a single partition and we'll just pick the first.

willbryant:
Yes, that's my understanding. So if two "anonymous" user events had the same exact timestamp with the same parameters, they would get combined to be one event (which may or may not be correct).

We could further reduce the likelihood of this occurring by also grouping by other fields. The origin of my particular list here is that that is the set of fields that goes into session_key & event_key, so it aligns with the uniqueness test we're going to run.

I suspect it's enough, but keen to hear others' thoughts.

willbryant:
Thanks for doc updates!

willbryant:
Incidentally can you ever get an audience trigger when the user_pseudo_id is null? I would have guessed that GA4 wouldn't be able to build an audience without giving the user a persistent identifier?

adamribaudo-velir:
> Incidentally can you ever get an audience trigger when the user_pseudo_id is null?

Good call. Pretty certain I read that when consent is denied, they can't be a part of a GA4 audience. 

One last point I have about this PR is that deduping as part of a `base` model chafes against certain dbt conventions. I think we're backed into this approach due to how BQ operates, but before the next release we may want to rename the model or otherwise add documentation to clarify the rationale.

adamribaudo-velir (approved):

willbryant:
Yeah, I would ideally like to address #80 and what I have in mind there would also address #33 - I'd like to made the "base" layer a single query on `events_*` and do the persisting-or-not in the next layer after that, with the same _table_suffix filtering mechanism in both the persisted daily and the non-persisted intraday/since-last-persisted queries.

Probably clearer in a PR than my explanation :)

PR #84, opened 2022-10-30
Utm content and term

willbryant:
## Description & motivation

Extract content & term UTM parameters too
    
Until now we've extracted 3 of the standard UTM tracking parameters (documented at https://en.wikipedia.org/wiki/UTM_parameters#UTM_parameters). This completes the set.

Our marketing agencies use these fields to track campaign creatives and paid keyword performance.

NB. I don't love how generic-sounding these column names are, but I've followed the pattern we've used so far. Perhaps we should be renaming them all back to `utm_*` to be clearer, but that isn't how they come through in event_params.

## Checklist

### DIFF
--- a/models/staging/ga4/stg_ga4__sessions_traffic_sources.sql
+++ b/models/staging/ga4/stg_ga4__sessions_traffic_sources.sql
@@ -21,24 +21,12 @@ set_default_channel_grouping as (
 session_source as (
     select    
         session_key,
-
-        (case
-          when FIRST_VALUE(source) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) is null then "(direct)"
-          else FIRST_VALUE(source) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) end
-        ) as session_source,
-
-        (case 
-          when FIRST_VALUE(medium) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) is null then "(none)"
-          else FIRST_VALUE(medium) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) end
-        ) as session_medium,
-
-        (case 
-          when FIRST_VALUE(campaign) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) is null then "(direct)"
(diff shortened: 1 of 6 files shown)


models/staging/ga4/stg_ga4__sessions_traffic_sources.sql
```diff
@@ -21,24 +23,14 @@ set_default_channel_grouping as (
...
-
-        (case 
-          when FIRST_VALUE(campaign) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) is null then "(direct)"
-          else FIRST_VALUE(campaign) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) end
-        ) as session_campaign,
-
-        FIRST_VALUE(default_channel_grouping) OVER (PARTITION BY session_key ORDER BY event_timestamp ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING) AS session_default_channel_grouping
+        COALESCE(FIRST_VALUE(source) OVER (session_window), "(direct)") AS session_source,
+        COALESCE(FIRST_VALUE(medium) OVER (session_window), "(none)") AS session_medium,
+        COALESCE(FIRST_VALUE(campaign) OVER (session_window), "(direct)") AS session_campaign,
```
adamribaudo-velir, line 28: I realize this issue existed before this PR, but isn't it problematic to assume that a null `campaign` value is `(direct)`? There is plenty of non-direct traffic that never has a campaign assigned. 

Seems like `(none)` would be more appropriate for an empty campaign value, but I might be missing something.

willbryant: Honestly I'm not too sure why we put values in here rather than leaving them NULL and letting the end user's viz tool take care of it, but I think (?) that the (none) and (direct) values are attempting to emulate what GA4's UI shows.
Screen Shot 2022-10-31 at 13 18 59
Screen Shot 2022-10-31 at 13 19 38

In that second screenshot we can see that there are separate (direct) and (not set) values, so looks like GA4 is being a bit more conditional about what it says - I assume the default value it puts in the campaign field depends on the source and/or medium.

More general question then, do we want to be trying to emulate that stuff, or should we leave it all NULL and let the end user figure out what they want to say?

willbryant: (referral) and (organic) seem a little different to me because they are actually based on some positive evidence - not just the absence of any other value. (referral) happens when there's an external, non-excluded referrer; organic is IIRC used with Google search referrals specifically.

I don't think they fill in anything at all for sessions that have no clear origin?

I guess I'd like it if the materialized tables (such as dim_ga4__sessions which uses this view) did not do any of this null-filling kind of thing, but we could do it in mart views. Then it's easy to keep composing more views on top of the materialized tables (for example, I'm probably about to have to do some extra work to find the first and last non-none session attribution per user, which would be an easy `FIRST/LAST_VALUE(.. IGNORE NULLS)` if null).

If we are keeping it, I don't have strong feelings about what the value should be, but the other fields mostly use (none), so would that be better than (not set)?

willbryant: You know, we're getting a bit off-topic here. I've pushed (none) to align with the others, how do you feel about that?

adamribaudo-velir (approved):

PR #86, opened 2022-11-01
Solution to create unique event key by redesign event_timestamp using event parameter & gtm timestamp variable.

00kennedyp:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,20 @@ packages:
 packages:
   - local: ../dbt-ga4
 ```
+
+
+## GTM configuration unqiue timestamp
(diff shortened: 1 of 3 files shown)


adamribaudo-velir:
@00kennedyp appreciate the PR, but we just had a robust discussion about this very issue in #74 and then ended up going with the approach in #83.

Your solution was considered, but you have to account for mobile hits that can't leverage GTM variables and also server-side generated events. The only way to guarantee there are no duplicate events is to dedupe them on the way into the staging models. Hope that makes sense.

adamribaudo-velir:
Also, feel free to open an issue first to discuss approaches before writing any code. Or chatting with any of us on DBT Slack or Measure Slack. Could save time and help to generate new ideas!

PR #87, opened 2022-11-02
Support deriving session properties from user_properties too

willbryant:
## Description & motivation

We have user_properties and derived_user_properties, and later I introduced derived_session_properties. But, there was no way to get a user_property value at a session level.

What I've found is that some user properties do slowly change over time, and so taking the very latest value from user_properties for the user means join queries show incorrect historic data changes.

This gives us the option to limit user_property data collection to session length. It happens to be a more natural data model for one of my user properties to think of it as a session level thing anyway; we use user properties rather than event params for that one because we want the GA4 behavior of applying user properties back to the start of the session for its built-in reporting.

## Checklist


adamribaudo-velir:
Great idea. I hope people spot this feature and use it.

adamribaudo-velir (approved):

PR #91, opened 2022-11-23
Fix Session Conversion unit test and update model name to _daily

adamribaudo-velir:
## Description & motivation

The `session_conversions` unit test broke with the recent changes in v2. In fixing that, I realized that it's a daily-partitioned table, so it would be more accurate to name it `stg_ga4__session_conversions_daily`

## Checklist
- [x ] I have updated the README.md (if applicable)


adamribaudo-velir:
Very simple change. Going to go ahead and merge.

Issue #97, opened 2022-11-25
Not enough resources for query planning


clemens7haar:
In the default configuation building the fct_ga4__pages model fails due to this message:

`Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex`

![image](https://user-images.githubusercontent.com/20857437/203935292-b674a9d4-07b3-4fe6-a751-c0878db076ff.png)

It's a really big database and I'm using the daily+streaming setting. After defining materialize table for base_ga4__events_intraday it works again, but I'm not sure if it's the best way to solve it. What would you recommend?

![image](https://user-images.githubusercontent.com/20857437/203936652-e96950e5-5917-442a-ba43-7c2d093597fb.png)

adamribaudo-velir:
Interesting. I know @dgitis is using this model on a large instance, but I believe he's only using the "streaming" option rather than "daily+streaming"

The only risk I can think of in materializing the intraday data as a table is freshness. You'll need to rerun dbt to see any newly streamed events. Seems like a solid work-around for now that doesn't require editing any package source files.

willbryant:
The cause of these errors is the number of CTEs and joins. You are more likely to run into these if you have a lot of "conversions" defined because the current implementation makes one CTE + join per "conversion". The size of the data shouldn't have any direct impact.

I fixed similar errors in other models in 746e949 and 300b926. I haven't run into this case myself as I don't use the pages model, so I have it disabled . But I think that in principle you should be able to apply a similar fix - count all the conversions in a single pass.

adamribaudo-velir:
I don't have a good way to test this, but if @willbryant is right, #98 should fix this

PR #98, opened 2022-11-27
Refactor page conversion count and unit test

adamribaudo-velir:
## Description & motivation
Responding to #97 

Refactored how page conversions are counted. Create `page_key` in the `stg_ga4__events` model rather than multiple times downstream. Should help with performance.

## Checklist

### DIFF
--- a/models/marts/core/fct_ga4__pages.sql
+++ b/models/marts/core/fct_ga4__pages.sql
@@ -16,13 +16,14 @@ with page_view as (
         event_date_dt,
         extract( hour from (select  timestamp_micros(event_timestamp))) as hour,
         page_location,  -- includes query string parameters not listed in query_parameter_exclusions variable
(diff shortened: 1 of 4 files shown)


willbryant:
I'm not really clear on the expected semantics of page_key with the hour component of the key, but I notice that the old code had a DISTINCT. Was the idea that the same event on the same page in the same hour (on the clock, not within a rolling 1h window) would only be counted as one conversion?

adamribaudo-velir:
Are you referring to this?

`conversion_{{ce}} as (
    select
        distinct (concat( cast(event_date_dt as string), cast(hour as string), page_location )) as page_key,
        sum(event_count) as conversion_count,
    from events
    where event_name = '{{ce}}'
    group by page_key
)`

That feels like possibly a mistake? Why distinct + group by? @dgitis can you clarify?

I think we would want a sum of all the conversions within the hour.

willbryant:
Yeah, that. I see, yeah seems redundant.

willbryant (approved):

PR #99, opened 2022-11-28
user_id mapping table that enables distinct user_id and user_pseudo_id marts

adamribaudo-velir:
## Description & motivation
Related to #62 

## Checklist


models/staging/ga4/stg_ga4__user_id_mapping.sql
```diff
@@ -0,0 +1,29 @@
...
+        event_timestamp 
+    from {{ref('stg_ga4__events')}}
+    where user_id is not null
+        and user_pseudo_id is not null
+),
+include_last_seen_timestamp as (
+    select 
+        user_id,
+        user_pseudo_id,
+        max(event_timestamp) as last_seen_timestamp
```
dgitis, line 14: If someone arrives at a site not logged in or cookied, they start the session with a `user_pseudo_id` but no `user_id`. They complete a few page_view events and then log in setting a `user_id`.  The navigate the logged in portion registering a few page_view events before logging out. Subsequent page_views will not have a `user_pseudo_id`. This is how the GA4 interface applies user IDs, but Google has always said that if you want to stitch this data together within sessions and across sessions (which is a situation that doesn't apply to this comment), then you should do that in BQ.

I'm willing to approve this PR as an improvement over what we have but I wouldn't call the `user_id` problems resolved with this PR. I think we need to get the maximum timestamp by `user_pseudo_id` and then attach that to the most recent, non-null `user_id` to completely resolve `user_id` issues within a session.

unit_tests/test_stg_Ga4__user_id_mapping.py
```diff
@@ -0,0 +1,43 @@
+import pytest
```
dgitis, line 1: Fix the capitalization of the file name.

models/marts/core/fct_ga4__pages.sql
```diff
@@ -19,7 +19,7 @@ with page_view as (
         page_key,
         page_title,  -- would like to move this to dim_ga4__pages but need to think how to handle page_title changing over time
         count(event_name) as page_views,
-        count(distinct user_key ) as users,
+        count(distinct user_pseudo_id ) as users,
```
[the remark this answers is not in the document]
dgitis: Alternately, switch this out to use the `user_id_or_user_pseudo_id` as seen in `fct_ga4__user_ids`.

dgitis (commented):
I'll want to run this code too, but these are my comments on reading through the code.

models/marts/core/fct_ga4__pages.sql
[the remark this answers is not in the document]
adamribaudo-velir: yea at a minimum, I think changing it to `devices` and going through to be consistent in the usage of `devices` vs. `users` makes sense.

models/staging/ga4/stg_ga4__user_id_mapping.sql
adamribaudo-velir: As far as I can tell, what's implemented is what you described in your last paragraph. This model looks for the most recent, non-null `user_id` for each `user_pseudo_id`

Are you saying this should be done session-by-session rather than across all events?

dgitis: Yes, you are right that it maps the last `user_id` to the `user_pseudo_id`. See my comment further down, though.

models/staging/ga4/stg_ga4__user_id_mapping.sql
```diff
@@ -0,0 +1,29 @@
...
+    from events_with_user_id
+    group by 1,2
+),
+pick_latest_timestamp as (
```
dgitis, line 18: This might be confusion on my part as I don't know these changes as well as you and it depends on how the `last_seen_timestamp` is used in subsequent models.

It is possible, though, that what you want for `last_seen_timestamp` is the last timestamp mapped to a `user_pseudo_id` regardless of the presence or absence of a `user_id`. This is to account for  sessions that extend after logging out.

However, it is possible that our session models are sufficient to account for these cases and that we want to preserve this as `user_id` focused. We'll need to communicate this.

adamribaudo-velir: Currently, the `last_seen_timestamp` field is only used within this model to pick a winner when multiple user_ids may have been associated with a single `user_pseudo_id` (which should be rare, but possible. Example would be a library computer where a single browser might see different users log in). 

So maybe `last_seen` is too vague and could be renamed to `last_seen_user_id_timestamp` to make it clear this isn't the last time we've seen the user's device. 

So going back to your example of a logout... the user's events after logout would be associated with their `user_id` in the `fct_user_ids` model which I think is the expected behavior.

adamribaudo-velir:
@dgitis renamed a few fields based on your feedback. lmk if you approve or have more feedback.

dgitis (approved):

rbarreca:
I think I'm running into the need for this. Am still new to dbt and working with GA4 data, but I was trying to find all sessions that occurred before a conversion. I have a user_id and therefore can create a hashed user_key, but the two sessions fact tables only expose user_pseudo_id. 

I have orders and user_id, but how do I join sessions_daily using the models?

PR #102, opened 2022-12-13
Use first event instead of first session_start

clemens7haar:
## Description & motivation

The dim_ga4__sessions model didn't contain all sessions, because sometimes the session_start event is missing. In this version it's based on the first event of a session. The issue is also described here: #69 

The stg_ga4__sessions_traffic_sources model needed some finetuning afterwards. For some sessions there was only a session_start or first_visit event tracked in my database and no further events. This caused NULL values for the traffic sources in the dim_ga4__session model.

## Checklist


adamribaudo-velir:
@dgitis I was under that impression as well, but I dug into a larger implementation and I see plenty of `session_start` and `first_visit` events with referrer and traffic source information. See image. Can you find any documentation saying otherwise? 
image

dgitis:
The project that I was working on used a GTM config with page_view events disabled and a separate page_view tag and the first_visit and session_start events never had page_referrer set but all other events were set.

I'll look into this a bit more.

adamribaudo-velir:
Shoot, maybe you're right. I was looking at `traffic_source_medium` etc which I think is actually user-leve. Here is source,medium, campaign which are null for `session_start` and `first_visit` 

image

dgitis:
Yeah, traffic_source.* is user-level.

dgitis:
I see two issues with looking at first page_view.

First, it puts us farther away from supporting mobile which uses screen_view.

Second, people can load a page in one session, walk away, come back for a second session, and generate events without any page_view. I was specifically looking at this scenario (among other things). 

I might have mis-remembered the page_referrer not being populated. I was looking at source, medium, default_channel_grouping, and page_referrer. I thought they were all not populated, but it looks like I was wrong with that detail. From what I saw, though, all events except for first_session and session_start gave the same attribution data as looking for the first event in a session where source was set. 

Querying first event where source is set requires an inefficient window function. Excluding first_session and session_start from a query with an `order by timestamp` is the more efficient option and it gave me in my testing the exact same results.

I'll check a few other projects, though.

willbryant:
I had to solve this problem in my own project and sort of wrote out the dbt-ga4 mechanism to do it.

My replacement is to use FIRST_VALUE(... IGNORE NULLS) to find the first event that does have UTM tracking parameters on it, without hardcoding the event names.

There is a catch, which is that multiple events in a session may have UTM parameters, and they won't necessarily all events will have the same set of UTM *parameters* (not just values).

That is, one might just have the first two or three out of source, medium, campaign, and content, and another might have all four.

It's highly undesirable to return a "torn read" where you see the source & medium from one event and the content and perhaps campaign from another, as then you see non-sensical values.

So, my actual code looks like this:

```
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN source END) IGNORE NULLS) OVER (session_window) AS source,
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN COALESCE(medium, '(none)') END) IGNORE NULLS) OVER (session_window) AS medium,
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN COALESCE(source_category, '(none)') END) IGNORE NULLS) OVER (session_window) AS source_category,
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN COALESCE(campaign, '(none)') END) IGNORE NULLS) OVER (session_window) AS campaign,
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN COALESCE(content, '(none)') END) IGNORE NULLS) OVER (session_window) AS content,
    FIRST_VALUE((CASE WHEN source <> '(direct)' THEN COALESCE(term, '(none)') END) IGNORE NULLS) OVER (session_window) AS term,
```

Plus I have some extra stuff for combining Firebase tracking things that I won't go into here.

(This is all just for extracting UTM parameters from individual events ourselves - none of this applies if you're using GA4's pre-delivered traffic source fields. But they don't return campaign which I need, and don't seem to align to the rules I want anyway.)

adamribaudo-velir:
Very helpful, @willbryant . Glad we won't make that mistake. 

Sounds like we're settled on using the first event with a `source` param, rather than looking for just `page_view`. I'll see if I can work on this over the weekend unless anyone beats me to it :)

adamribaudo-velir:
This has been updated in https://github.com/Velir/dbt-ga4/pull/102/commits/406f4618c17456a5af4995839e5175e3a8a8ccaf based on the discussion here. Can someone review?

dgitis (dismissed):
I believe that `stg_ga4__sessions_first_session_start_event` gets orphaned with this change. At least, the first project that I checked indicates that deleting the reference to it orphans the model.

Before deleting that model, be sure to check whether `dim_ga4__sessions` triggers any query complexity errors. As long as that's not a problem, delete the orphaned model and merge these changes.

PR #111, opened 2023-01-25
Allow users to override default_channel_grouping

adamribaudo-velir:
## Description & motivation
Implementation of `adapter.dispatch` within the `default_channel_grouping` macro to allow end-users to override this macro and use their own channel mapping. Instructions were added to the README:

```
By default, this package maps traffic sources to channel groupings using the macros/default_channel_grouping.sql macro. This macro closely adheres to Googls recommended channel groupings documented here: https://support.google.com/analytics/answer/9756891?hl=en .

Package users can override this macro and implement their own channel groupins by following these steps:

Create a macro in your project named default__default_channel_grouping that accepts the same 3 arguments: source, medium, source_category
Implement your custom logic within that macro. It may be easiest to first copy the code from the package macro and modify from there.
Overriding the package's default channel mapping makes use of dbt's dispatch override capability documented here: https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch#overriding-package-macros
```

## Checklist
- [ x] I have run `dbt test` and `python -m pytest .` to validate exists tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ This package uses `pytest` as a method of unit testing individual models. More d
 
 By default, this package maps traffic sources to channel groupings using the `macros/default_channel_grouping.sql` macro. This macro closely adheres to Googls recommended channel groupings documented here: https://support.google.com/analytics/answer/9756891?hl=en .
 
-Package users can override this macro and implement their own channel groupins by following these steps:
+Package users can override this macro and implement their own channel groupings by following these steps:
 - Create a macro in your project named `default__default_channel_grouping` that accepts the same 3 arguments: source, medium, source_category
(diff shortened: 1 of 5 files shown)


adamribaudo-velir:
Ran into what may be a bug in dbt-core. Not sure. Slack thread and MRE here: https://getdbt.slack.com/archives/CU4MRJ7QB/p1674690798050729?thread_ts=1674654356.581109&cid=CU4MRJ7QB

willbryant:
What was the outcome of that thing about a possible DBT bug?

adamribaudo-velir:
oh, turned out to just be an ambiguity in the documentation. Users need to override the adapter-specific macro, so in our case `default__default_channel_grouping()` rather than `default_channel_grouping()`

It's all set and works as I've tested it.

dgitis (approved):

Issue #117, opened 2023-01-31
Use cluster on event_name field of base_ga4__events in order to reduce costs


franzoni315:
Hello!

I am beginning to try this package and this is awesome! I see that the `base_ga4__events` table is already partitioned by time (`event_date_dt` field), but there are many views that depend on this table and filter it only by `event_name` (for instance `stg_ga4__event_video_start`). In order to reduce costs of BigQuery, a simple solution is to use the `cluster_by` option of the dbt's `config` statement:

```
{{
    config(
        materialized = 'incremental',
        incremental_strategy = 'insert_overwrite',
        partition_by={
            "field": "event_date_dt",
            "data_type": "date",
        },
        partitions = partitions_to_replace,
        cluster_by=['event_name']
    )
}}
```

Maybe there could be an option, in order to allow even other clustered fields. Currently, in my company's datawarehouse, we also cluster by user id, and stream id.

adamribaudo-velir:
Great suggestion! Do you want to open a PR?

franzoni315:
Great!! I am eager to see the performance gains! I may post the results here after the approval of the PR.

PR #119, opened 2023-02-01
Added event_name clustering to base event model

adamribaudo-velir:
## Description & motivation
 
As noted in #117, we can speed up the filtering by event name by using BQ clustering (documented here: https://cloud.google.com/bigquery/docs/clustered-tables ). Performance gains are seen when tables are larger than 1GB.

## Checklist


adamribaudo-velir:
I'll merge this in the next 24h unless anyone has any objections.

dgitis (approved):

dgitis:
I thought that I had seen this and then confused it with #120 . Sorry for the delay.

Issue #115, opened 2023-01-29
Evaluate ingestion-time partitioning and partition copy available in dbt 1.4.0 


adamribaudo-velir:
https://medium.com/teads-engineering/bigquery-ingestion-time-partitioning-and-partition-copy-with-dbt-cc8a00f373e3

dbt 1.4.0 now supports ingestion-time partitions (as opposed to column type partitions). It seems as there are some cost savings to be had when partitions are very large (> 1M rows per partition). It's not immediately clear to me if this benefits our package, but seems worth considering.

Another feature they introduced is the `bq copy` feature for handle `insert_overwrite` incremental merges which seems relevant. 
![image](https://user-images.githubusercontent.com/71573504/215348992-2fba8553-8599-48ae-96d4-4b69a697269b.png)

PR #120, opened 2023-02-01
Use copy_partition strategy for base event incremental model to save on processing time and cost

adamribaudo-velir:
## Description & motivation

As mentioned in #115, dbt 1.4.0 released the ability to use BQ's 'copy' command as opposed to a 'merge' command to update incremental models. This can save on both processing time and cost. In my own testing of merging in 1.4M rows, I saw the following results:

Status quo:
```
3.27 GB billed

03:19:22  1 of 1 START sql incremental model dbt_dev_aribaudo.base_ga4__events ........... [RUN]
03:20:03  1 of 1 OK created sql incremental model dbt_dev_aribaudo.base_ga4__events ...... [SCRIPT (3.3 GB processed) in 41.09s]
03:20:03
03:20:03  Finished running 1 incremental model in 0 hours 0 minutes and 41.98 seconds (41.98s).
```
With copy_partition:

```
03:17:02  1 of 1 START sql incremental model dbt_dev_aribaudo.base_ga4__events ........... [RUN]
03:17:40  1 of 1 OK created sql incremental model dbt_dev_aribaudo.base_ga4__events ...... [None (0 processed) in 38.13s]
03:17:40
03:17:40  Finished running 1 incremental model in 0 hours 0 minutes and 39.06 seconds (39.06s).
```

So not only did I shave 2 seconds off the processing time, but I was billed for 0 bytes instead of 3.8GB. Seems like a no brainer, given the ease of configuration and its availability in dbt-bigquery.

## Checklist

### DIFF
--- a/models/staging/ga4/base/base_ga4__events.sql
+++ b/models/staging/ga4/base/base_ga4__events.sql
@@ -9,8 +9,7 @@
             incremental_strategy = 'insert_overwrite',
             partition_by={
                 "field": "event_date_dt",
-                "data_type": "date",
-                "copy_partitions": true
+                "data_type": "date"
             },
             partitions = partitions_to_replace,
             
--- a/models/staging/ga4/base/base_ga4__events.sql
+++ b/models/staging/ga4/base/base_ga4__events.sql
@@ -10,8 +10,10 @@
             partition_by={
                 "field": "event_date_dt",
                 "data_type": "date",
+                "copy_partitions": true
             },
             partitions = partitions_to_replace,
+            
         )
     }}
 {% else %}
@@ -22,7 +24,9 @@
             partition_by={
                 "field": "event_date_dt",
                 "data_type": "date",
+                "copy_partitions": true
             },
+            
         )
     }}
 {% endif %}


willbryant:
Does this actually work? The docs say it's only for time_ingestion_partitioning mode.

dgitis:
I tried to replicate this both with and without time_ingestion_partitioning on dbt Cloud (making sure to set the environment to dbt 1.4) and it errored every time but dbt was being odd. Maybe dbt Cloud caches the environment; maybe it was just broken. My local dev environment is having some problems as well, so I gave up there, but this needs to be tested and confirmed because the docs don't match the PR and I suspiciously can't get any combination to work.

adamribaudo-velir:
Hi, where in the docs does it say that the bq copy method requires ingestion time partitions? I don't believe that's the case.

The example in the medium post provides a config without ingestion time partitions

{{ config(
    materialized = 'incremental',
    incremental_strategy = 'insert_overwrite',
    partition_by = {
      "field": "day",
      "data_type": "date",
      "copy_partitions": true
    }
) }}



Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

charles-astrafy:
image

adamribaudo-velir:
ah, thanks. I see that now in the dbt documentation. So weird that it's not mentioned in the Medium post.

When I run this (using dynamic partitions, not static incremental) I get the expected results:
image

But it looks like this doesn't work with the `static_incremental_days` setting enabled. Get the following compilation error:

```
19:44:48  Compilation Error in model base_ga4__events (models\staging\ga4\base\base_ga4__events.sql)
19:44:48    'str object' has no attribute 'strftime'
19:44:48
19:44:48    > in macro bq_static_insert_overwrite_sql (macros\materializations\incremental_strategy\insert_overwrite.sql)
19:44:48    > called by macro bq_insert_overwrite_sql (macros\materializations\incremental_strategy\insert_overwrite.sql)
19:44:48    > called by macro bq_generate_incremental_insert_overwrite_build_sql (macros\materializations\incremental_strategy\insert_overwrite.sql)
19:44:48    > called by macro bq_generate_incremental_build_sql (macros\materializations\incremental.sql)
19:44:48    > called by macro materialization_incremental_bigquery (macros\materializations\incremental.sql)       
19:44:48    > called by model base_ga4__events (models\staging\ga4\base\base_ga4__events.sql)
19:44:48
19:44:48  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```
So this might not be as valuable as I had imagined. I'll leave the PR open for now to consider options.

PR #123, opened 2023-02-03
Modified section "Custom Parameters"

Sgiostra-BitBang:
In explanation of how to use the **Custom Parameters**, the doc shows that as project must be use **ga4:** (key  of second level of section **var:**).
But to use the _Custom Parameters_ with the models that I defined in the my project (that obivously uses the package ga4) I must said to _ga4_ that the model _\<my_model\>_ associated to that _Custom Parameters_ (_\<my_model\>\_custom_parameters_) is in **\<my_project\>** and not in **ga4**.

I propose this modify because i think that this aspect needs a more clear explanation.

- [X] I have updated the README.md (if applicable)
- [X] I have added tests & descriptions to my models (and macros if applicable)
- [X] I have run `dbt test` and `python -m pytest .` to validate exists tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ vars:
 ```
 ### Custom Parameters
 
-Within GA4, you can add custom parameters to any event. These custom parameters will be picked up by this package if they are defined as variables within your `dbt_project.yml` file using the following syntax:
+Within GA4, you can add custom parameters to any event. These custom parameters will be picked up by this package if they are defined as variables of **your project** within your `dbt_project.yml` file using the following syntax:
 
 ```
 [event name]_custom_parameters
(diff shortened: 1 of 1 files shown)


dgitis:
This change to the docs would mislead anyone adding custom parameters to the built-in staging models that should be nested under a `ga4:` block.

This is a dbt knowledge issue rather than a package issue, so I'm not sure that we should  change the docs at all. 

However, if we were to change the docs to make it easier for people new to dbt starting out using the package, then I would prefer that we create a new section on **Adding Custom Events** where we cover how to add staging models, different ways of supporting custom parameters on staging models, and warn there that vars and model configs nested under a `ga4` node won't affect custom events defined in the project and that you should set those under a `project id` node.

Sgiostra-BitBang:
Hi @dgitis, 
I think your idea is a good idea. By adding a new section to the document, we can add this new information without changing the original intent of the document.

adamribaudo-velir:
Created Issue #135 and will close this PR.

PR #126, opened 2023-02-06
Multi site

dgitis:
## Description & motivation
Multi-site support is a frequently requested features. Currently, to use the package in a multi-site installation, you need to override the base model and then override several downstream models. In doing this, you make upgrading the package difficult as you need to implement new features in each model that you override.

Additionally, because of dbt's one-model to one-table approach, there's no way of dynamically creating multiple base models which greatly complicates multi-site support.

For this reason, I chose to support multi-site installations by creating base templates that can be used to add as many sites as needed and a array variable to list the sites. When the array variable is present, the `base_ga4__events` model unions all of the configured base templates before passing the unioned data on to the downstream models as normal.

In addition, those downstream models are designed to detect the existence of the array variable, and add a `stream_name` to the model (derived from a seed mapping `stream_id` to `stream_name`) allowing end users of the package to query all sites or a subset of sites as they desire. This has the added effect that only when we publish rare package changes to `base_ga4__events` will multi-site users need to update their models.

## Checklist
- [ y] I have run `dbt test` and `python -m pytest .` to validate exists tests


dgitis:
@adamribaudo-velir, I got pytest working but I'm having problems with `test_base_ga4__events.py`. 

It's creating and running the tests fine, but the data is not loading into BQ. As you wrote the upload_json_fixture, I was hoping that you would have some insight into why the tests aren't working. I suspect that I'm missing something stupid.

I'll also want to implement some multi-site tests to the model (the current test only tests single-site), but I'm pretty confident that once I have the single-site test finally working, the multi-site will be quite easy.

3v-dgudaitis (commented):
Close: resolved by #150

Issue #106, opened 2023-01-13
Time on Page


dgitis:
I frequently get customization requests to add a time on page metric. I always push back with performance objections, but, because the clients who ask for time on page all value it very highly, they have so far always insisted.

Right now I have two requests open: one is for a small non-profit that isn't going to exceed the free tier with the request and the other is for a high-volume publisher who doesn't care how much it costs.

I'm thinking of two general design patterns for implementing this both of which involve adding a variable to the dbt_project.yml.

The first is to create an "enable_non_performant_metrics" variable with a list value of consisting of metrics that require window functions ['time_on_page', 'exits', etc...] and then enable those metrics as configured.

The second is to create an "enable_non_performant_utitlities" variable with a true-false value that adds session_first_event_id, session_last_event_id, and next_page_view_event_id (possibly last_page_view_event_id, I'm not sure if we needit) to events possibly using one of the methods below if set to true.

The session_first_event_id would use select the first event that is not session_start or first_visit as only those two events fire before the config tag loads page_referrer and other similar data. I've had to debug some session issues recently and I'm confident that this works as intended.

@willbryant and I talked about doing these sorts of lookups between events_* tables and base_ga4___events so that the window functions don't go wild looking across partitions. This goes against dbt best-practices, so I understand any resistance, but we at least have a good reason to do so.

Another possibility is to create a stg_ga4__window_functions table that runs parallel to (or just after) stg_ga4__events with a hard-set window. I'm not sure that this would work, but the idea is to maintain the `stg_ga4__events` dynamic querying of partitions while also allowing the use of window functions (when building but now when running dynamic queries) that don't search all partitions.

Unfortunately, the first, clearly easier method doesn't really work as a first step towards realizing the second, much more comprehensive method.

I'd like to hear what other think.

adamribaudo-velir:
Can you describe why time on page is a "non-performant" metric? In my limited understanding, there is the `engagement_time_msec` field which magically does this for us.

dgitis:
It does time on site.

```
select
  sum(engagement_time_msec) and time_on_site
from table
group by session_key
```

If you have a better way of doing time on page, I'd love to hear it; but here's what I think needs to happen.

```
select
  timestamp_diff(next_page_view_timestamp, event_timestamp, second) as time_on_page
from(
  select
    event_timestamp,
    lead(event_timestamp) over (partition by session_key order by event_timestamp asc) as next_page_view_timestamp
[...]

PR #127, opened 2023-02-09
Time on page

dgitis:
## Description & motivation
The current implementation of `total_time_on_page` in `fct_ga4__pages` uses calculates time on page as `sum(engagement_time_msec)`. However, `engagement_time_msec` is the engagement time elapsed since the previous event. It works for summing across the span of a session, but not in this context. This results in landing pages getting ridiculously low `total_time_on_page` numbers because they are effectively calculating the time in microseconds between the config tag loading and the `page_view` event firing.

This PR adds a calculation to the `stg_ga4__event_page_view` model that calculates the difference between the the current page_view and the next page_view in the session.

The `fct_ga4__pages` model gets `total_time_on_page` (replacing the current, flawed definition) and `average_time_on_page_denominator` fields. The denominator is required for the calculation to work when grouping by anything as the average function works when working in SQL by throwing out null denominators, but the math breaks when averaging the averages in an external data vis tool so we need to provide a field with the null denominators filtered out.

The advantage of this approach is that it should match existing UA data more closely. We may also want to implement a total engagement time on page metric, bit it is not something that GA4 exposes and it a little more complicated than `total_time_on_page` as calculated in this PR.

## Checklist
- [ y] I have run `dbt test` and `python -m pytest .` to validate exists tests


adamribaudo-velir (changes requested):
Hey, sorry but 2 issues that I think need to be addressed:

1) Biggest one I added as a comment to the linked issue: https://github.com/Velir/dbt-ga4/issues/106#issuecomment-1385328882 . The event timestamps are not guaranteed to be in order and will represent the batch time, not the time of collection. 

From Simo Ahava: "This means that two things become impossible to reliably analyze: 1) When any given event actually happened, and 2) what the sequence of events was."

I feel pretty funny about adding in a feature that we know won't be correct. What is your take here?

2) We've avoided window functions in many cases because it, from what I can tell, necessitates a scan of all base partitions. I'd ask that we add the timestamp_diff to a staging layer above `stg_ga4__event_page_view` so that processes accessing the `page_view` event won't trigger that scan every time. I think conceptually this makes sense anyyway, since we're enriching the model in a novel way.

adamribaudo-velir:
To respond to both of my concerns, what about calculating ENGAGED time on page as:

```
select
  page_location,
  sum(engagement_time_msec) and time_on_site
from stg_ga4__events_user_engagement
group by page_location
```

This will only capture engaged time, not the entire amount of time, but it's more aligned with where GA4 is headed anyway (focusing on engaged time) and removes the need for window functions.

dgitis:
I'm not sure that I understand engagement time in GA4. I think `engagement_time_msec` is the elapsed engagement time since the last time engagement time was set. Look at this session:

| event_name      | page_location                 | event_date | event_timestamp  | ga_session_id | engagement_time_msec |
|-----------------|-------------------------------|------------|------------------|---------------|----------------------|
| first_visit     | https://caretjuice.com/       | 20230206   | 1675713351611201 | 1675713352    |                      |
| session_start   | https://caretjuice.com/       | 20230206   | 1675713351611201 | 1675713352    |                      |
| page_view       | https://caretjuice.com/       | 20230206   | 1675713351611201 | 1675713352    |                      |
| page_view       | https://caretjuice.com/       | 20230206   | 1675713359344891 | 1675713352    | 4842                 |
| scroll          | https://caretjuice.com/       | 20230206   | 1675713367457533 | 1675713352    | 3608                 |
| user_engagement | https://caretjuice.com/       | 20230206   | 1675713373527494 | 1675713352    | 11074                |
| page_view       | https://caretjuice.com/about/ | 20230206   | 1675713374505334 | 1675713352    |                      |
| scroll          | https://caretjuice.com/about/ | 20230206   | 1675713386531572 | 1675713352    | 7079                 |
| user_engagement | https://caretjuice.com/about/ | 20230206   | 1675713455466712 | 1675713352    | 36227                |

So, total engagement time is 4842 + 3608 +11074 +7079 + 36227= 62,830 milliseconds.

How do we calculate engagement time on page, though? Only one page_view event has an engagement time. I presume that it is the elapsed engagement time from the beginning of the session.

Subtracting page_view 1 timestamp from page_view 2 timestamp yields 7,734 milliseconds against an engagement time of 4,842 milliseconds which would be page 1's `total_engagement_time_on_page`. That makes sense that the engagement time would be less than the timstamp diff. 

Subtracting pag_view 2 timestamp from page_view 3 timestamp yields 15,161 milliseconds.  If we sum the two `engagement_time_msec` values between page_view 2 and page_view 3, we get 14,682 milliseconds for page_view 2's `total_engagement_time_on_page`.

The page_view 3's calculation would be all remaining engagement time. For comparsing we subtract page_view 3's timestamp from the last event timestamp and get 80,961 milliseconds and the elapsed engagement time is 43,306 milliseconds.

So we need to sum the `engagement_time_msec` values for all events following but not including the initial `page_view` up to and including those values in the next `page_view`. 

That's not just a simple `sum(engagement_time_msec)` like it is in the session table.

dgitis:
One possibility for your concerns about the batching of timestamps is to calculate using the event_timestamp, but allow configurations that replace the event_timestamp with a custom timestamp of your choosing.

dgitis:
As far avoiding the impact of window functions crossing partitions are concerned, we have two options as far as I can see and one extra thing to consider.

Because the `lead()` function looks forward, it might not scan older partitions. I suspect that this is not true because date partitioned tables in BigQuery seem to have no date logic coded into them and they could be arbitrary-number-partitioned tables as far as BQ is concerned. But we should check and be sure.

For the options, here's what I see.

1. We move any window functions into `base_ga4__events`
2. We move window functions to after the staging models

If we could save the result of `_dbt_max_partition` to its own tables, we could pass that value to the staging views and limit partition scanning when building the package, but then we would need to reset the views with a query without `_dbt_max_partition` for ad-hoc queries that follow. 

This would definitely involve the over-riding of a dbt Core macro and it might possibly best be an PR against dbt Core.

dgitis:
I had some time to think about engagement time vs. time on site and I think I've got an idea that will work and avoid window functions.

In `stg_ga4__events` we add what I'll call an `engagement_key` column after the `session_key` is generated from the following case statement:
```
case
  when event_name = 'page_view' then concat(session_key, page_referrer)
  else concat(session_key, page_location)
end as engagement_key
```
In `stg_ga4__event_page_view` we add the following:

```
with engagement as (
  select
    engagement_key,
    sum(engagement_time_msec) as page_engagement_time_msec
  from stg_ga4__events
  group by engagement_key
)
select * from page_view_data
left join engagement using(engagement_key)
```

adamribaudo-velir:
I like it where you're going with this. Can you please make this a new staging model outside of `stg_ga4__events_page_view`? Maybe something like `stg_ga4__page_engaged_time`

Given how identical the other `event` models are, I don't think anyone would expect this logic to live in `stg_ga4__events_page_view`

dgitis:
I think the previous version to was better than the current version. I suspect the fat group by clause `fct_ga4_pages` is going to be a lot more expensive than the previous version.
```
 page_engagement as (
    select
        page_view.* except(page_engagement_key),
        sum(page_engagement_time) as total_engagement_time,
        countif(page_engagement_time is not null) as average_engagement_time_denominator
    from {{ ref('stg_ga4__page_engaged_time') }}
    right join page_view using (page_engagement_key)
    group by 1,2,3,4,5,6,7,8,9,10
),
```

And it means that people wanting to build a `fct_ga4__event_page_view` model have to know about the engagement time table rather then just see it in the `stg_ga4__event_page_view` model.

adamribaudo-velir:
@dgitis can you document the .yml file what the `page_engagement_key` represents? I want to make sure I'm understanding why you generate a key from `page_referrer` in some cases and `page_location` in others and it seems like that should be documented in the package.

dgitis:
This is my description of the `page_engagement_key` field in `stg_ga4.yml`.

> This field is used to assign engagement_time_msec to the correct page. While the GA4 documentation claims that a user_engagement event that sets engagement_time_msec should fire when navigating to a new page, the data shows that this is not always the case. When a user_engagement event does not fire, the engagement_time_msec parameter is assigned to the next page_view. This engagement time should be credited to the previous page, so for page_view events this field uses the session_key and page_referrer as the key while all other events use the session_key and page_location. 

I also considered the possibility that the subsequent page_view would be batched and `engagement_time_msec` might be set on a different event. To test this, I ran this query and didn't find any event batches with page_view events that set `engagement_time_msec` on anything other than the page_view event.

```
with prep as (
SELECT  
  event_name,
  concat(session_key, event_timestamp) as batch_key,
  case
    when engagement_time_msec is not null and event_name = 'page_view' then 'page_view has engagement_time'
    when engagement_time_msec is not null and event_name != 'page_view' then 'other event has engagement time'
    else null
  end as engagement_status,
  case
    when event_name = 'page_view' then 1
    else 0
  end as is_page_view

FROM `project.dataset.stg_ga4__events`
where event_date_dt = '2023-02-16'
), batch as (
  select
    batch_key,
    max(is_page_view) as has_page_view,

  from prep  
  where engagement_status = 'other event has engagement time'
  group by 1  
)
select
  sum(has_page_view) as total_batches_with_page_view_and_engagement_on_other_event
from batch
```

adamribaudo-velir:
Great, thanks. I'm following you. One last question: how do we later join in `stg_ga4__page_engaged_time.sql` to get the page of interest? 

If the user's 2nd `page_view` is on `/page-2.html` and the `page_referrer` is for `/page-1.html` then the `page_engagement_key` allows you to aggregate your metrics against `/page-1.html` which is what you want. But if you want to build a report out of this, you need a dimension to aggregate off of that's human-readable. 

I think you might need a `page_engagement_page` field in that model as well? That make sense?

dgitis:
The previous version had `engagement_time_on_page`, or something similar, in the `stg_ga4__event_page_view` model and then I build the `fct_ga4__pages` model from there. You asked me to remove it to keep the event models consistent.

I will renew my preference for restoring that version as adding engagement time to that event model is easier for end users to understand and use and it would provide the human-readable parameter that you are requesting.

There is one issue that I've become aware of. In sessions where someone visits /page1 then browses around and visits /page1 again, this implementation would sum the engagement time on both pages and then apply that total amount twice, once for each page view.

I'm thinking about fixes in `fct_ga4__pages` that involve concatenating the session_key and page_key and then changing the `average_engagement_time_denominator` from `countif(page_engagement_time is not null)` to `count(distinct case when page_engagement_time is not null then session_key else null end)`.

I'm not sure how to fix the `total_engagement_time` without using a  window function.

dgitis:
This is as close to perfect as I can get it. The one final issue that I can't figure out without using a window function is described in the `core.yml` file's description for the `avg_engagement_time_denominator` column.

> Use avg_engagement_time_denominator to calculate the average engagement time, which is derived by dividing the sum of total engagement time by the product of the sum of the denominator and 1000 to get the average engagement time in seconds (average_engagement_time = sum(total_engagement_time_msec)/(sum(avg_engagement_time_denominator) *1000 )). The denominator excludes page_view events where no engagement time is recorded for the page_location within a session. However, it includes subsequent page_view events to a page_location that has previously recorded a page_view event in the same session, even if the subsequent event has no recorded engagement time.

To get the calculation perfect, we need to know if the page is an exit page. 

I considered briefly whether I could use the technique that I used in `stg_ga4__page_engaged_time.sql` where the `page_engagement_key` had already been assigned to the `session_key + page_referrer` and then I reassigned it to the `session_key + page_location` to also calculate whether a page is also exit page without using a window function (it's used there to assign engagement time to the correct page). 

It may be possible, but I can't see how to get past issues like multiple visits to the same page in a session with only the last one being an exit.

adamribaudo-velir (changes requested):
@dgitis `fct_ga4__pages.sql` contains a compilation error and then some database errors. I tried to work through it a bit (compilation error is just some text hanging over from a git conflict) but got stuck on the database errors. Look like some new columns aren't grouped properly. 

I know we had previously discussed removing the hour-grain from `fct_ga4_pages`. Want to do that here? Or save that for later?

dgitis:
Sorry, I'll deal with the database error. I probably forgot to check the errors when I merged changes from main.  Was it a query complexity error? I've gotten those from too many group by clauses in the pages model and actually removed hour from a client site to fix it. I guess that's one more good reason to delete the hour grouping on the pages table.

adamribaudo-velir:
Nah, not query complexity. Just a few issues related to the number of columns compared to the GROUP BY clause. I solved 1 and another popped up and figured you'd be best equipped to sort through it.

adamribaudo-velir:
@dgitis I'm able to run it now, thanks. I do, however, get about 50% null values in fct_ga4__pages for engaged_time_msec. Is that expected?

image

I also think some documentation in the .yml files would be helpful to explain `page_engagement_key`. But if the null values are expected then I suppose we can merge.

dgitis:
I'd left the hour in on the page_key in `stg_ga4__events`. That's fixed now.

adamribaudo-velir (approved):

PR #130, opened 2023-02-14
Dispatch the unnest_key macro

### LINKED ISSUE #129: Event parameters can be recorded as multiple data types, our unnest_keys() macro doesn't account for this
Flagged by a user in Measure Slack here: https://measure.slack.com/archives/CHC9RDUSG/p1676045661507979

From Simo Ahava: "If the value is numeric it ends up in the int (or double) value columns, if it’s something else then in the string column"

Currently, the unnest_keys macro only looks for values in 1 field (meaning 1 data type). 

Other BQ users have decided to use COALESCE to find the first non-null value. We should think through the ramifications.

adamribaudo-velir:
## Description & motivation
As noted in #129, there are cases where event params can appears as multiple data types and package users may need custom logic to respond. This PR dispatches the `unnest_key` macro so that package users can implement their own logic for unnesting event params. See docs here for more on macro dispatching: https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch

Closes #129 

## Checklist

### DIFF
--- a/macros/unnest_key.sql
+++ b/macros/unnest_key.sql
@@ -1,13 +1,14 @@
 -- Unnests a single key's value from an array
 
 {%- macro unnest_key(column_to_unnest, key_to_extract, value_type = "string_value", rename_column = "default") -%}
+    {{ return(adapter.dispatch('unnest_key', 'ga4')(column_to_unnest, key_to_extract, value_type, rename_column)) }}
+{%- endmacro -%}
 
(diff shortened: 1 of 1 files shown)


adamribaudo-velir:
Going to merge in the next 24h unless there are any objections.

willbryant:
Seems fine to me.

BTW I think an alternative offering would have been to support eg. "any_value" as a pseudo-type that does the COALESCE-strings. But people can just implement that themselves now.

PR #132, opened 2023-02-15
Model determining whether a page_location and session combination reached 90% scroll

adamribaudo-velir:
## Description & motivation
While working on updates to the `pages` fact table, I realized that when simply counting the number of scroll events, you end up with an amount proportional to the granularity of tracking. For instance, if a user tracks scrolls at every percentage, you get a large count, but if they track scrolls at 25%/50%/75%/100% they get a small count. 

A better method would be to count the number of scroll events that reach a certain threshold (I'm suggesting 90% based on the OOB scroll tracking Google provides). This staging model measures whether a 90% scroll occurred per date, `page_location` and `session_key`.

Future work can incorporate this into fact tables. 

## Checklist

### DIFF
--- a/models/staging/ga4/stg_ga4.yml
+++ b/models/staging/ga4/stg_ga4.yml
@@ -83,4 +83,5 @@ models:
         tests:
           - not_null
           - unique
-
+  - name: stg_ga4__scroll_90_pct
+    description: Determines whether an individual page_location during an individual's session triggered a >=90 scroll or not.
--- a/models/staging/ga4/stg_ga4__scroll_90_pct.sql
+++ b/models/staging/ga4/stg_ga4__scroll_90_pct.sql
@@ -1,0 +1,9 @@
+-- Determines whether >=90% scroll was reached per date, page_location and session_key
+select
+    event_date_dt
+    , page_location
+    , session_key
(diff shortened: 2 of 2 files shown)


dgitis:
Are you sure that this is necessary? [Google's docs](https://support.google.com/analytics/answer/9216061?hl=en) claim that the default scroll event only fires at 90 percent. 

I've seen errors in GA4 documentation before, but I just want to be certain that we need to do anything.

adamribaudo-velir:
The default for the automatic `scroll` event only fires at 90%, right, but Velir's implementations with clients we typically replace that with a GTM-fired event tracking 25/50/75/90 because clients want more granularity than just 90% scroll.

It's probably not the most impactful model, but it makes downstream marts more resilient to implementation differences and makes it more explicit when you want to count 90% vs. other percentages.

dgitis (changes requested):
I'm a little uncomfortable with this as it seems to assume customization. 

I'll approve it because it works with the default configuration of the scroll event as well as the custom scroll event.

However, the `fct_ga4__pages` model has some scroll metrics. That model needs to be changed to use this model (rather than `stg_ga4__event_scroll` as it does now) so that it works the same way for default and custom scroll configurations.

adamribaudo-velir:
I should also pull in the `event_key` as well. Noting that here.

PR #134, opened 2023-02-20
Add session_number field to dim_ga4_sessions

adamribaudo-velir:
## Description & motivation

- Renamed `ga_session_number` to `session_number` which I think will be more intuitive to users
- Added `session_number` to dim_ga4__sessions as well as an `is_first_session` flag

## Checklist

### DIFF
--- a/models/marts/core/dim_ga4__sessions.sql
+++ b/models/marts/core/dim_ga4__sessions.sql
@@ -38,6 +38,8 @@ with session_first_event as
         device_web_info_browser,
         device_web_info_browser_version,
         device_web_info_hostname,
+        session_number,
+        session_number = 1 as is_first_session,
         traffic_source_name,
         traffic_source_medium,
         traffic_source_source,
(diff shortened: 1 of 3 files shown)


dgitis (approved):
I'm happy to approve this but I think it leads to some questions that we should consider collectively. 

Our default MO has been to use the BigQuery GA4 naming conventions. I think that we should consider whether we want to rename fields and under what conditions.

For example, the traffic_source_* fields are poorly named. We should rename them to user_* because, when working with data vis resources who likely don't have knowledge of the raw data export, the end consumer can easily confuse trafiic_source.source with source. Particularly when the former field is populated throughout the session while the latter field is only populated on the first batch or events.

As a standard, then we should do all of our naming transformations either in the base_ga4__events model or the stg_ga4__events model. I'm fine with either, but let's choose to be consistent after this.

adamribaudo-velir:
Yea I agree. I've come around to thinking that we should smooth over the rough edges of Google's obtuse naming conventions. Users can always refer back to the original `base` model to see any renaming that's occurring.

To date, I believe we've consistently handled renaming in the `base` model but let me know if I'm wrong. Agree that we should rename original field names once and in a single, consistent location.

PR #137, opened 2023-02-23
Fix engaged sessions

ivan-toriya-precis:
## Description & motivation

Fix session_engaged event parameter. The value should be `int_value`, not `string_value`

## Checklist

### DIFF
--- a/models/staging/ga4/base/base_ga4__events.sql
+++ b/models/staging/ga4/base/base_ga4__events.sql
@@ -129,7 +129,7 @@ renamed as (
         {{ ga4.unnest_key('event_params', 'ga_session_id', 'int_value') }},
         {{ ga4.unnest_key('event_params', 'page_location') }},
         {{ ga4.unnest_key('event_params', 'ga_session_number',  'int_value') }},
-        (case when (SELECT value.string_value FROM unnest(event_params) WHERE key = "session_engaged") = "1" then 1 end) as session_engaged,
+        {{ ga4.unnest_key('event_params', 'session_engaged',  'int_value') }},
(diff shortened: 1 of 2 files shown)


adamribaudo-velir:
@ivan-toriya-precis in the projects I have access to, I see `session_engaged` set as a string:
image

That's also what I see in this blog post here: https://www.ga4bigquery.com/sessions-dimensions-metrics-ga4/

Do you have any documentation from Google that is more concrete? Thanks for the PR!

ivan-toriya-precis:
@adamribaudo-velir 
It's not documented or I just couldn't find in anywhere.
For the client I'm working with, it's in `int_value`, also double-checked with another one:
image

But, for some, I've found them **sometimes** in the `string_value` where platform = "WEB"
Weird, I'll try to investigate more about it.

ivan-toriya-precis:
Yes, I can look into that, but probably starting next week 👍

ivan-toriya-precis:
@adamribaudo-velir ready for review

adamribaudo-velir (approved):
Looks good! Thanks so much.

eliesg:
for the enterprise client i work with, `session_engaged` can be in both int_value and string_value in a WEB only dataset:

Plein_écran_03_03_2023_07_23

So i coalesce both values:
```
(select COALESCE(value.string_value, cast(value.int_value as string)) from UNNEST(event_params) where key = 'session_engaged') as session_engaged,
```

I see you've done something similar in base_ga4__events, but with a CASE WHEN. Anyway it was just to add up on that topic :)

PR #145, opened 2023-03-04
Remove extra ga4 folder

### LINKED ISSUE #142: Remove models/staging/ga4 folder and move staging models under /models/staging?
When I started this package, I put the models under /models/staging/ga4 because typically I put staging models under a folder named after the source of data. This, however, creates confusing situations where users need to select package models using this selection criteria: `ga4.staging.ga4.MODEL` and this project config:

```
models:
  ga4:
    staging:
      ga4:
        recommended_events:
          stg_ga4__event_sign_up:
            +enabled: true
          stg_ga4__event_share:
            +enabled: true
```

It also seems unnecessary because ALL data in this package is sourced from ga4. It would be much cleaner to remove the /models/staging/ga4 folder. But this would be a breaking change for some folks who reference those paths.

Thoughts?

adamribaudo-velir:
## Description & motivation
Closes #142 

Removing the `ga4` folder under the `staging` folder as it likely causes confusion to users who need to specify model configs under the `models` section of `dbt_project.yml`.

Existing users will need to update their `dbt_project.yml` as a part of this change. 

## Checklist


dgitis (approved):
I will be adding two Issues, one on renaming individual fields and another on our model naming structure. Whatever we decide, I think we should get all of the renaming pain over with in the same release.

PR #150, opened 2023-03-11
Multi property support using copy command

adamribaudo-velir:
## Description & motivation
This PR adds support for multiple GA4 properties using a pre-hook macro on the `base_ga4__events` model that copies event shards from multiple datasets into a single combined dataset. The pre-hook is enabled when the `property_ids` variable is set. 

The major advantage of this approach is that data copied using the `clone` keyword is copied for free. This is as opposed to alternative approaches that have been suggested using a `union` operator across tables which would incur costs. One disadvantage is that the target dataset will be populated with 1 table per property per day. This is only visual clutter as the tables can be queried and otherwise managed using [wildcard tables queries](https://cloud.google.com/bigquery/docs/querying-wildcard-tables). 

Known limitations:
- Only supports `daily` frequency
- Supports full refresh and `static_incremental_days` incremental strategy, does not support "dynamic" incremental refresh. 

This PR adds a dependency on dbt_utils to help dynamically pull the available table shards with the `dbt_utils.get_relations_by_pattern` macro.

## Checklist


dgitis:
I like this approach with the pre hook and combine_property_data macro.

We're also going to want to add `stream_id`, or replicate the stream name feature I proposed in #126, to all of the models in the marts folder to simplify the filtering of different streams.

adamribaudo-velir:
@dgitis done! good catch. I think this is ready to go as long as we're ok with it not handling intraday/streaming yet. I'd be interested in releasing it and getting feedback before over-engineering it.

eric-systematik:
> @dgitis done! good catch. I think this is ready to go as long as we're ok with it not handling intraday/streaming yet. I'd be interested in releasing it and getting feedback before over-engineering it.

If we only deal with one GA4 property, would it still handle the streaming feature? I know for my clients, data freshness is essential.

adamribaudo-velir:
Yes. The multiple property feature only engages when multiple property IDs are entered into the 'property_ids' variable.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

dgitis (approved):

Issue #151, opened 2023-03-13
Add 'Push Notifications' default channel


gblazq:
The GA4 team has added a default channel grouping named 'Mobile Push Notifications' (https://support.google.com/analytics/answer/9756891). Given that push notifications are a common channel for content publishers, it would be a nice feature to add so that the package and the GA4 behaviour match.

At the same time, I think 'Push Notifications' is a much better channel name than 'Mobile Push Notifications', because most push notification services are not restricted to mobile devices and can send notifications to browsers on any device (for example, OneSignal is one of the most used push notification services and their own docs suggest using 'push' as the medium, no matter the device https://documentation.onesignal.com/docs/links#utm-parameters)

I have a branch ready to create a pull request with those two changes, but I had doubts on whether the channel name change was appropriate or not. Given that the default channel grouping macro does not exactly match the behaviour of GA4, I think the name change is justified, but it's also something that any end user can handle on their own. Any thoughts on this?

adamribaudo-velir:
I'm ok with the name change based on your comments. I don't think it's reasonable for anyone to expect an exact 1:1 match between BigQuery vs. GA4 Native attribution results, so if we can provide clearer labels then that's a plus. 

Let's leave this open for a few days to see if there are other comments, though. Thanks!

adamribaudo-velir:
Closed as part of #152

PR #152, opened 2023-03-13
Add push notifications channel

gblazq:
## Description & motivation
The GA4 team has added the 'Mobile Push Notifications' channel group (see https://support.google.com/analytics/answer/9756891). This is an important channel for news outlets and other publishers that use push notifications.

I have renamed the channel from 'Mobile Push Notifications' to just 'Push Notifications' because 'push' is a common medium value for browser push notification services too.

This PR would close issue #151 with the channel name change.

## Checklist

### DIFF
--- a/macros/default_channel_grouping.sql
+++ b/macros/default_channel_grouping.sql
@@ -50,6 +50,8 @@ case
     then 'Audio'
   when {{medium}} = 'sms'
     then 'SMS'
+  when REGEXP_CONTAINS({{medium}}, r"(mobile|notification|push$)") or {{source}} = 'firebase'
+    then 'Push Notifications'
   else '(Other)' 
 end 
 
--- a/unit_tests/test_macro_default_channel_grouping.py
+++ b/unit_tests/test_macro_default_channel_grouping.py
@@ -14,6 +14,8 @@ cn.bing.com,
 43things.com,
 alibaba,
 alibaba,cpc
+,push
+firebase,
 ,
 """.lstrip()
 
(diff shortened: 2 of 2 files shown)


dgitis (changes requested):
Thanks for fixing this up for everyone.

The [GA4 docs](https://support.google.com/analytics/answer/9756891?hl=en) give the following rules for the Mobile Push Notifications channel.

> Medium ends with "push"
>OR
>Medium contains "mobile" or "notification"
>OR
>Source exactly matches "firebase"

We want to default to matching these rules so you'll need to move 'push' into a separate regex to match the ends with rule (rather than push anywhere).

Additionally, `Push Notifications` should be  `Mobile Push Notifications`.

gblazq:
The regex matches the rules in the docs (`push$` ensures that `push` is only matched at the end of the string). As for the name, I had doubts and explained the reasons in issue #151. I have no problem with the required change, just wanted to point there because there's a comment in favour of the rename.

I should have set the PR as a draft before having more feedback on #151 to avoid duplicate discussions, sorry about that!

adamribaudo-velir:
@dgitis yea I think he's all set on the regex using `push$`. I don't have a really strong opinion on the default channel grouping name but I think if we need a tie breaker, we should go with what Google recommends. @gblazq do you mind making that 1 change? Then I think we can merge.

dgitis:
Sorry, I missed the `push$`. 

I also didn't see the issue where you had discussed _Push Notification_ versus _Mobile Push Notification_. I wouldn't approve a deviation from GA4 without a good reason, but it sounds like you have a decent reason for the change. 

I'll withdraw my change request.

dgitis (approved):

Issue #161, opened 2023-03-24
Multi-site faulty incremental logic


dgitis:
The `is_incremental` macro only works on incremental models. It does not work in the pre-hook of an incremental model. As a result, the `combine_property_data` macro always full refreshes.

Instead we should be using this:

```
{% if not flags.FULL_REFRESH %}
  -- incremental logic
{% endif %}
```

There's a fix staged in the multi-site-enhancements branch.

Issue #160, opened 2023-03-24
Multi-site time-outs


dgitis:
I'm not sure what we can do here, but running a full-refresh on 4 months of data across 5 sites with low-to-moderate traffic levels times out. 

It's not a huge amount of data. Maybe we recommend manually cloning in the event of a full refresh. 

I also exclude the `base_ga4__events` from full refreshes by default and only enable refreshing that table manually. Maybe we should make this a default behavior and add instructions on how to refresh

dgitis:
So it appears that setting the `+full_refresh: false` model config does not effect the `not flags.FULL_REFRESH` block in the `combine_property_data` macro. I guess that makes sense since it is not a flag we are referencing but rather a model setting.

Right now, I think the solution is to set up a manual procedure for cloning into the `var('dataset')` directory when you need to refresh the data. Then we would need to suppress any cloning on a full refresh (refreshing the base table from the clone folder). 

The incremental logic would all remain the same.

PR #162, opened 2023-03-27
Multi site enhancements

dgitis:
## Description & motivation
This PR bundles two improvements to the multi-site functionality.

**combine_property_data macro**

Switch to use `if not should_full_refresh()` rather than the `is_incremental()` macro as the is_incremental does not work in pre-hooks.

Additionally, this allows users to set the `base_ga4__events` config to use `+full-refresh: false` in the model config block as cloning on full-refreshes is prone to timing out.

This resolves issue #161. 

**friendly stream names**

Added support to map stream_ids to stream names via a seed file.

Because you can have multiple streams in a single property, this feature required an extra config parameter.

I know when I previously set this up for a different multi-site concept, Adam had requested that I implement the mapping via `dbt_project.yml` variables rather than using a seed. Using the seed works and is easy to implement.  Using a variable requires nested Jinja lists which then need to be converted into a SQL model. It's a pain to implement and I lack the time to do it right at the present.

I suggest we merge this working version and then enhance it later when people have the time to improve on it.

**documentation**

Added documentation recommending an increase to the BigQuery timeout settings when using multi-site to circumvent issue #160. 

**This PR is based off of the Time on page PR. That PR should be merged first.**

## Checklist


adamribaudo-velir:
is this a branch from the time on site branch? There seem to be 21 files changed.

dgitis:
Correct @adamribaudo-velir, there's a note at the bottom of the PR's description to merge the time-on-page PR first.

adamribaudo-velir:
@dgitis I merged in the 'time on page' branch but still getting a conflict with this branch. 

Could you separate out the 'stream rename' and 'combine_property_data macro' fix for #161? They are very different contributions. 

I'm a little skeptical that the 'stream rename' should be a part of the package - it seems like a BI layer concern in that it simply re-labels data to be human-readable. Similar to how you would use the BI layer to convert '20230329' to 'Wednesday March 29th'

dgitis:
I'm going to close this and reopen with a cleaner PR.

PR #166, opened 2023-03-29
base_select macro and combine property enhancement

dgitis:
## Description & motivation

**combine_property_data macro**

Switch to use `if not should_full_refresh()` rather than the `is_incremental()` macro as the is_incremental does not work in pre-hooks.

Additionally, this allows users to set the base_ga4__events config to use +full-refresh: false in the model config block as cloning on full-refreshes is prone to timing out.

This resolves issue https://github.com/Velir/dbt-ga4/issues/161.

**base_select macro**

Switch `base_ga4__events` and `base_ga4__events_intraday` to use a macro to set the contents of their select statements in order to make them easier to keep in sync and maintain.

**documentation**

Added documentation recommending an increase to the BigQuery timeout settings when using multi-site to circumvent issue https://github.com/Velir/dbt-ga4/issues/160.

## Checklist


eliesg:
hey @dgitis, I think that's a great idea to have a base_select macro, so that users like me can change how some event_params are selected, right?

I've started to use dbt for an enterprise client and i've taken a lot of inspiration from this package. In our implementation, several event_params can be recorded in different data types (see https://github.com/Velir/dbt-ga4/issues/129). Which means that in my base_ model, I sometimes need to do:
```
        (select COALESCE(value.string_value, cast(value.int_value as string), cast(value.double_value as string)) from UNNEST(event_params) where key = 'my_event_param') as my_event_param,
```
Otherwise, I would lose a lot of values, if I selected only `value.string_value` for example.

To add more complexity, we have several GA4 properties (several sub-domains for example) that we use with dbt. And the implementation for each of these GA4 properties can be different. In one property, `ga_session_id` is recorded as a STRING in BigQuery; in another property, it's recorded as an INT. Do you think I would be able to handle that, with the base_select macro?

adamribaudo-velir (approved):

PR #176, opened 2023-04-08
Fix: Multi property sources from source database, not target database

adamribaudo-velir:
## Description & motivation
Previously, the `combine_property_data` macro used the target database in the `dbt_utils.get_relations_by_pattern` function to find tables to copy over. We needed to set the `database` parameter to `var('project')` to make sure it looked in the correct database. This was only an issue for users who use different databases for raw data and modeled data. 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,20 +278,18 @@ Overriding the package's default channel mapping makes use of dbt's dispatch ove
 
 # Multi-Property Support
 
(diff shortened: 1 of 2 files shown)


adamribaudo-velir:
This is a pretty simple fix that's blocking some production work. Will merge in the next 48hrs unless there's any objection.

TannerHopkins (commented):
Thanks for fixing this!

PR #180, opened 2023-04-14
Update .gitattributes

ivan-toriya:
Infinitesimal GitHub cosmetics update


dgitis (approved):
From what I understand, what this does is that it adds the SQL files to the list of languages in GitHub, so instead of showing up as 100 percent Python in GitHub, it will show a mix of SQL and Python.

image

Issue #163, opened 2023-03-28
Multi-site: user_pseudo_id fails unique check


dgitis:
The user_pseudo_id fails the unique check when used in multi-site.

The user_pseudo_id needs to be concatenated with the stream_id to pass this check.

dgitis:
@adamribaudo  and @willbryant 

What do you two think about adding a `user_pseudo_key` to `stg_ga4__events` vs. simply replacing the `user_pseudo_id` with `to_base64(md5(concat(user_pseudo_id,stream_id)))`?

The first is consistent with what we have done in the past but also requires a bunch of changes to downstream models. The second is less destructive of reports built off of the package and easier to implement.

I could add an `original_user_pseudo_id` field for debug purposes if we go the second route.

adamribaudo-velir:
I'm leaning towards the first suggestion (`user_psuedo_key`). I have a feeling that modifying the value of the `user_pseudo_id` column will cause a lot of confusion for package users who might be troubleshooting and would reasonably expect that value to be exactly what Google sets. 

This might be a good opportunity to create some more consistent naming and use the column name `device_key` so that the hierarchy of entities becomes:
- `user_id`
- - `device_key`
- - - `session_key`
- - - - `event_key`

Good catch.

dgitis:
I'm assuming that `device_key` is what I was referring to as `user_pseudo_key`. I don't hate that name, but I immediately started thinking about device lookup tables and wondered why we would need those. 

How about `client_key`? It's send as the cid (for client ID) in the collection API, so there's some precedent. Client is also more accurate than device since you can have multiple clients on one device and the cid/user_pseudo_id corresponds to a client rather than a device.

adamribaudo-velir:
I like it! Let's go with `client_key`

willbryant:
Yep, sounds good. user_pseudo_key or client_key IMHO. Definitely agree not user_pseudo_id.

PR #181, opened 2023-04-14
Update user_pseudo_id to client_key

TannerHopkins:
## Description & motivation
There was a [recent issue raised](https://github.com/Velir/dbt-ga4/issues/163) for `user_pseudo_id` failing unique checks for multi-property implementations of the package. A decision was made to replace `user_pseudo_id` with `client_key` throughout the package, with `client_key` being a hash of `user_pseudo_id` and `stream_id`.
This PR accomplishes the planned work laid out in that issue, substituting `client_key` for `user_pseudo_id` throughout the package. 

A few notes:
- `user_pseudo_id` was only kept in the base_select.sql macro
- Whenever both `user_pseudo_id`  and `stream_id` were used in a group by statement, I substituted out _both fields_ for `client_key` to avoid redundancy
- I had to rename a few models that included user_pseudo_id in the name. I'm assuming this would cause breaking changes for some users of this package, so I'm curious if the new release that would include this PR would need to be a new major version (i.e. 4.0.0)?

## Checklist


TannerHopkins:
Just pushed one more commit to include this description for client_key whenever it is mentioned in YML files throughout the package:
```description: Hashed combination of user_pseudo_id and stream_id```
Feel free to suggest a better description, or let me know if we don't need this description everywhere. I at least wanted to replace the old description for `user_pseudo_id` from one of the YML files, and thought it might be a good idea to just add this everywhere while I'm at it since this new field isn't native to ga4.

File: models/marts/core/fct_ga4__sessions_daily.sql
Comment (dgitis) [inline models/marts/core/fct_ga4__sessions_daily.sql:1]: Revert `stream_id` and `group by` and re-add `stream_id` to the `join_metrics_and_conversions` cte.

File: models/marts/core/fct_ga4__user_ids.sql
Comment (dgitis) [inline models/marts/core/fct_ga4__user_ids.sql:1]: Revert `stream_id` and `group by`.

**Comment by TannerHopkins** [comment]
> Thanks for all of your hard work here. I was literally sitting down to resolve this myself. The PR looks good except for one issue affecting several files.
> 
> The `stream_id` is needed for multi-site installations and GA4 properties that combine app and web data to segment streams when reporting.
> 
> Fix that issue and then I'd like to run this on some test data before approving just because it touches a lot files.

Thanks for calling that out! I had wondered if removing that field would be an issue or not, was hoping it would be mentioned in PR review if so. For context, I haven't used this package a whole lot just yet - trying to get it set up for a very large multi-property implementation though (hundreds of properties) and @adamribaudo  mentioned this issue might be a blocker.

I'll fix this now!

**Review by dgitis — APPROVED** [review]
That looks good @TannerHopkins. Thanks for helping out.

File: macros/base_select.sql [hunk 1]
```diff
@@ -42,6 +43,7 @@
     event_server_timestamp_offset,
     user_id,
     user_pseudo_id,
+    to_base64(md5(concat(user_pseudo_id, stream_id))) as client_key,
```
adamribaudo-velir, line 46: @TannerHopkins we do all other key creation in `stg_ga4__events`. Could we move that there to a CTE early in the model?

PR #187, opened 2023-04-21
multi-site support for streaming frequency

dgitis:
## Description & motivation

This PR adds streaming support for multi-property installations.

I could have unified the `events_` and `events_intraday_` code in the `combine_property_data` macro a little more writing both to `events_*` tables because, as things are currently written, daily and streaming frequencies are mutually exclusive and saved a couple lines of code. However, it is likely that supporting the daily+streaming frequency will require separation of code and it is easier to debug when they write to  `events_` and `events_intraday_` code.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Example:
 ```
 vars:
   ga4:
-    frequency:"daily+streaming"
+    frequency: "daily+streaming"
 ```
 
 # Connecting to BigQuery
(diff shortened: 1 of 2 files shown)


adamribaudo-velir:
Look good. I didn't think it would be that easy. I guess `daily+streaming` would be hard because you'd have to juggle the fact that the streaming data is eventually merged into the daily data.

adamribaudo-velir (approved):

Issue #182, opened 2023-04-14
Combining events and events_intraday data where GA4 batch export exceeds quota


dgitis:
A GA4 property with both batch and streaming exports populates the intraday tables in almost real time and then the batch tables are populated, hopefully overnight, and the equivalent streaming partition gets deleted.

The batch export, however, is limited to 1,000,000 events per day. When a batch export exceeds that 1 million event threshold, the basic events partition is not created and the intraday partition that holds the streaming data remains in place.

Our current approach to combining the tables bolts on everything in the intraday table. It would be better to combine these tables in a more robust way as unioning the intraday partitions to the events data results in all intraday partitions getting reprocessed.

One possibility is to not distinguish between `events_*` and `events_intraday_*`. The issue with this approach is that I have heard, but not witnessed, that there can be a period of time where equivalent partitions exist on both the `events_*` and `events_intraday_*` tables.

I'd love to hear other ideas for how to combine older intraday tables in such a way as they do not need to be reprocessed.

dgitis:
@willbryant and @adamribaudo-velir, I'd like your opinions on this issue. 

I think it might be best for us to remove the `frequency` variable and the ``base_ga4__events_intraday` model entirely and treat all `events_*` and `events_intraday_*` tables the same.

Removing the distinction between those tables would resolve this issue. It might simplify the code. And it would make the package easier to use.

It's not the sort of thing that I'd want to change without discussion first.

willbryant:
Keen, please see https://github.com/Velir/dbt-ga4/issues/80 - specifically my proposed resolution would get easier with the change you propose, I think?

adamribaudo-velir:
I agree that handling the intraday table as a special condition in so many places is brittle and I'd be fine dropping the 'frequency' variable if we can find an elegant solution. I just wish I had a better handle on the variations that we need to support. From what I understand, there are 4:

1. Daily only
2. Daily + Streaming. The intraday table is dropped and converted to a daily shard
3. Daily + Streaming when some days have >1M events and no daily shard is written
4. Streaming only 

I don't have any examples of 3 & 4 to work with which makes me uneasy in producing a solution.

Based on what I know now, what a solution may look like is this:

- Select from daily and intraday tables into our `base` model and depend on our dedupe clause to remove duplicate rows
- Require users to use 'static incremental days' setting to ensure any day with partial data is reprocessed

[...]

PR #188, opened 2023-04-24
Refactor daily & streaming support

adamribaudo-velir:
## Description & motivation

Previously, this package supported daily + streaming (aka intraday) exports in a relatively naïve way by incrementally processing 'daily' data and then unioning the 'streaming' data as a view in the `stg_ga4__events` model. This can cause issues when tables are rotated from 'intraday' to 'streaming' as noted in #80 and in other cases noted in #182. 

This update ensures that daily & streaming data are handled using the same method: dats is loaded incrementally into the base model and deduped before being loaded into `stg_ga4__events`. 

As a byproduct, this PR removes support for 'dynamically' incremental processing based on the last date found in a table using `_max_partition`. Instead, the `static_incremental_days` variable will need to be used which will ensure that data is reprocessed as the intraday table changes throughout the day.

## Checklist


jice-lavocat:
👍 

Any missing Todos regarding this PR @adamribaudo-velir ? 
I could try to help since I'm looking at the package these days.

macros/combine_property_data.sql
```diff
@@ -16,22 +16,21 @@
 
     {% for property_id in var('property_ids') %}
         {%- set schema_name = "analytics_" + property_id|string -%}
-        {%- if var('frequency', daily) == 'streaming' -%}
+            -- Copy daily tables
```
jice-lavocat, line 19:
```suggestion
            {# Copy daily tables #}
```

macros/combine_property_data.sql
```diff
@@ -16,22 +16,21 @@
...
             {%- set relations = dbt_utils.get_relations_by_pattern(schema_pattern=schema_name, table_pattern='events_intraday_%', database=var('project')) -%}
             {% for relation in relations %}
                 {%- set relation_suffix = relation.identifier|replace('events_intraday_', '') -%}
                 {%- if relation_suffix|int >= earliest_shard_to_retrieve|int -%}
                     CREATE OR REPLACE TABLE `{{var('project')}}.{{var('dataset')}}.events_intraday_{{relation_suffix}}{{property_id}}` CLONE `{{var('project')}}.analytics_{{property_id}}.events_intraday_{{relation_suffix}}`;
                 {%- endif -%}
             {% endfor %}
-        {%- else -%}
+            -- Copy intraday tables
```
jice-lavocat, line 27:
```suggestion
            {# Copy intraday tables #}
```

jice-lavocat (changes requested):
SQL comments with Jinja code 'minification' (`{-` and `-}`) ) were inadvertently causing the CREATE commands to be disabled, leading to undesired outcomes.

I posted a suggestion to remove them (change `--` into `{#` ).
Your PR seems to run fine except from that.

--

I also applied your changes to my branch ([multiple GA property config](https://github.com/jice-lavocat/dbt-ga4/tree/streaming-daily)). I don't know what would be the best to merge the two though.

models/staging/base/base_ga4__events.sql
```diff
@@ -1,61 +1,49 @@
...
-        from {{ source('ga4', 'events_intraday') }}
-        where cast( _table_suffix as int64) >= {{var('start_date')}}
-    {% else %}
         from {{ source('ga4', 'events') }}
         where _table_suffix not like '%intraday%'
         and cast( _table_suffix as int64) >= {{var('start_date')}}
+    {% if is_incremental() %}
+        and parse_date('%Y%m%d', left(_TABLE_SUFFIX, 8)) in ({{ partitions_to_replace | join(',') }})
     {% endif %}
+),
```
dgitis, line 28: I think we need to add something like this below code to prevent data double-counting in the period between when GA creates a new events partition and before it deletes the equivalent events_intraday partition. @willbryant reported some issues related to this.

```
max_source_daily as (
  select max(event_date) as max_date
from source_daily
```
and then further down in the source_intraday block

```
where cast( _table_suffix as int64) >= {{var('start_date')}}
and cast(_table_suffix as int64) != max_source_daily.max_date
```
I think it has to be a does not equal condition here as this PR is supposed to also solve old events_intraday partitions that should be processed when the batch processing does not happen due to exceeding the 1 million event limit. So, we want to process old events_intraday partitions, just not the max partition when it equals the events max partition.

This assumes two things:
1. That the max events partition is always complete
2. Google will delete the intraday partition before the next day's processing.

I think 1 is a safe assumption.

For 2, it will still be possible for this to happen.

Day 1: partitions exist for `events_Day 1 - 1` and `events_intradayDay 1 - 1` and the package processes Day 1 - 1 as desired
Day 2: partitions still exist for  `events_Day 1 - 1` and `events_intradayDay 1 - 1` (equivalent of Day 2 -2) and Day 1 data gets doubled

I think this is unlikely, certainly much less likely than not suppressing the max partition from source_daily on source_intraday.

adamribaudo-velir: @dgitis what's the harm  in letting the dedupe logic at the end of `base_ga4__events` catch this, if there are duplicate events across daily & intraday?

dgitis: The user ltv data comes in its own column that's not deduped so that's not an issue. 

I think it would be fine. It might cost a little extra processing, but we don't need everything to be perfect as long as it's functional.

We might run into a problem in the future if Google adds any processed data to `event_params`. 

Is the `google / cpc` traffic coming through as `google / organic` problem only a problem on intraday? Or is it there for batch as well?

If it's a problem on both then I think we can ignore my comment.

dgitis (approved):
I've tested this on my local machine and approved this but I don't currently have a test project that uses multi-site and I don't currently have a test project that combines batch and streaming.

I would feel better about approving this PR if someone had tested these to the point where, if no one has, I might just create another test project to fulfill those scenarios for this PR and future ones.

Otherwise, this is so much cleaner. I love it.

jice-lavocat:
As far as I am concerned, I have tested it on a multi-site project, in which some are in streaming, some in batch, but none of them use both mechanisms. 
Therefore I can't comment on other potentiel issue such as the recent discussion about potential duplication (I would also assume that if we are sure we deduplicate later in the process, then this is fine).

adamribaudo-velir:
Thanks for the feedback, all. Merging.

PR #189, opened 2023-04-26
add user_id and session_number to session tables

dgitis:
## Description & motivation

The way that our tables are currently architected, we can't get a daily user count and returning users without querying the client and user tables that aren't partitioned. 

In order to let people get these metrics, this PR adds `user_id` and `session_number` to the `fct_ga4__sessions_daily` and `fct_ga4__sessions` tables. 

This will allow users to get new and returning users by running queries like `count(distinct case when session_number = 1 then ifnull(user_id, client_key) else null end) as new_users` or `count( distinct ifnull(user_id, client_key)) as users` against the sessions table and not query all data for all time.

## Checklist
- [n/a ] I have updated the README.md (if applicable)


models/marts/core/fct_ga4__sessions.sql
```diff
@@ -4,18 +4,20 @@ select
     client_key,
     session_key,
     stream_id,
+    user_id,
```
adamribaudo-velir: In the case of a multi-day session, user_id could be null on day 1 and have a value on day 2. This would result in 2 rows with the same session_key. Should we take the max(user_id) here as well?

adamribaudo-velir (approved):

adamribaudo-velir:
@dgitis take a look at my updates and feel free to merge if they look good.

dgitis:
I seem to have failed to commit changes that added `max(user_id)` to both models and that addressed the column name error. My apologies.

PR #195, opened 2023-05-04
Fix test on user_id_or_client_key in fct_ga4__user_ids model

### LINKED ISSUE #191: Uniqueness test failing in fct_ga4__user_ids
We're seeing the `unique_fct_ga4__user_ids_user_id_or_client_key` test fail. It's only happening for one `client_key`, but that `client_key` has 3 different `stream_id` values. 

Perhaps this was an unintended effect of swapping out `user_pseudo_id` for `client_key` in [this PR](https://github.com/Velir/dbt-ga4/pull/181)? Maybe there could only be one `user_pseudo_id` per `stream_id` before, but now there can be multiple?

Regardless, I did find it interesting that it is grouping by `user_id_or_client_key` and `stream_id` in `fct_ga4__user_ids`, but then testing for uniqueness on `user_id_or_client_key` by itself:
```
select
    user_id_or_client_key,
    stream_id,
    max(is_user_id) as is_user_id,
    min(first_seen_timestamp) as first_seen_timestamp,
    ....
from user_id_mapped
group by 1, 2
```

I'm not sure if the fix would be to update the uniqueness test to take into account `stream_id` as well, or possible stop grouping by `stream_id` and instead take the min/max, similar to what is being done with `max(is_user_id) as is_user_id` above.

adamribaudo-velir:
## Description & motivation
Closes #191

@TannerHopkins noticed that the uniqueness test for `fct_ga4__user_ids` can fail when joining together multiple data streams as it checks only the `user_id_or_client_key`. The `client_key` value already takes the `stream_id` values into account, but `user_id` does not. This means that sharing a `user_id` across streams would invalidate the test (despite being a valid outcome).

The test has been updated to incorporate `stream_id` in the uniqueness check. 

## Checklist

### DIFF
--- a/models/marts/core/core.yml
+++ b/models/marts/core/core.yml
@@ -28,9 +28,3 @@ models:
         description: Hashed combination of user_pseudo_id and stream_id
         tests: 
           - unique
-  - name: fct_ga4__user_ids
-    description: Fact table with aggregate metrics at the level of the user_id when one is present, otherwise at the device level (as indicated by the client_key). Metrics are aggregated from fct_ga4__client_keys.
-    columns:
-      - name: user_id_or_client_key
-        tests: 
-          - unique
--- a/models/marts/core/fct_ga4__user_ids.yml
+++ b/models/marts/core/fct_ga4__user_ids.yml
@@ -1,0 +1,10 @@
+version: 2
+
+models:  
+  - name: fct_ga4__user_ids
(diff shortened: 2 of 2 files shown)


adamribaudo-velir:
@TannerHopkins this should be all set if you want to retest

TannerHopkins (approved):
Working for me now - approved!

PR #196, opened 2023-05-05
Last non-direct session attribution

### LINKED ISSUE #110: Last non-direct attribution
I have this as a client requirement so I'll be working on this soon. Either I'm missing something, or it's a surprisingly complicated problem so I wanted to give others a chance to weigh in.

I see three basic approaches to this problem:

1. We do last non-direct in a View that sits on top of the session marts**and only lookup the last non-direct session within the partitions being queried**. This is efficient but has some obvious data quality flaws.
2. We lookup the last non-direct session while building the session marts. The data quality in this case will be much better but it will be horribly inefficient on large sites as it requires a lookup that we may not be able to constrain by partitions.
3. We offload last non-direct to the client and add `last_non_direct_source`, `last_non_direct_medium` and `last_non_direct_campaign` parameters and use those to define `last_non_direct_channel_grouping` and filter all of those parameters up to the session models. This suffers from some data quality issues in that cookies can be deleted and it only works at the client level (not the user level) without using lookups that compromise any efficiency gains.

The second option is the one that I find superficially most attractive, but the more I think about the details of that implementation, the less practical I think it is.

If we were to lookup the sessions table on user_key to find the last non-direct session, we would create a circular dependency with the sessions table referencing the sessions table. 

We could store the last non-direct session attribution values in the user table. 

At first I thought this would be enough, but then I realized that all of that users' sessions would change to the most recent non-direct session when a user has a new non-direct session. 

So, we could store the last non-direct session values in the user table, and then lookup that value when building the session tables.

This would mostly work when running daily builds. But what happens when a user has two different non-direct sessions in a day (with direct sessions in between)? And also what happens when we do a full-refresh on the data warehouse?

Given the problems with these two approaches, the only solution I can see is to create a new table for storing non-direct sessions (`timestamp`, `user_key`,  `last_non_direct_source`, `last_non_direct_medium`, `last_non_direct_campaign`, and `last_non_direct_channel_grouping`). This will still be a pretty beefy table on large sites, but it will work.

I feel like there's a more elegant solution that I'm missing.

If anyone can think of a better approach, I'd love to hear of it. Otherwise, what approach do you prefer? Did I miss any issues or advantages? @adamribaudo-velir @willbryant

adamribaudo-velir:
## Description & motivation
Closes #110 

The following model has been added which provides last non-direct session attribution: `stg_ga4__sessions_traffic_sources_last_non_direct_daily`

This model is similar to `stg_ga4__sessions_traffic_sources_daily` but provides the following new fields:
- session_source_last_non_direct
- session_medium_last_non_direct
- session_source_category_last_non_direct
- session_campaign_last_non_direct
- session_content_last_non_direct
- session_term_last_non_direct
- session_default_channel_grouping_last_non_direct

Each field contains the traffic source values from the last session with a non-direct `session_source` value. The effect can be seen in the graphic below where the '(none)' channel grouping is replaced with 'Organic Search'

image

A unit test has been created which mimics this behavior and validates the results. 

## Checklist


dgitis:
I'm thinking the solution here is to build a minimal session table partitioned by day that contains session data possibly with no window functions and certainly with no lookback period. I'll call it `int_ga4__sessions_daily`.

If we aim for no window functions, we also add a partitioned `page_view` event table, `fct_ga4__event_page_view`.

On top of these two tables will build a daily sessions table `dim_ga4__sessions` or `dim_ga4__sesssions_daily` that uses a configurable lookback window (30 days is a good default, but this is really a business decision, so it should be in a variable. Since we have the lookback window configured here, we also reconcile cross-partition sessions in this query when hitting `int_ga4__sessions_daily` and get first_last page_view data from `fct_ga4__event_page_view` with only a one-day lookback while also resolving last_non_direct.

The goal of this architecture is to have all window functions hitting smaller tables.

Another issue that I've ran into when I set this up for a client on an old fork of the project is sessions with only `first_visit` and `session_start` events. Because we filter these events out from our session attribution, we end up failing to include those session keys in our session attribution queries, so when we join attribution data later, there's no data to join against.

We need downstream models to change null to direct or we need to decide to that we want to filter out these sessions. I think changing null to direct to be safer because it's hard to anticipate how people will use GA4. In my client's case, removing those sessions was the right call, but I'm not ready to make that generalization.

models/marts/core/dim_ga4__sessions_daily.sql
dgitis, line 1: I'm won't block merging this PR because of this, but I feel like a better naming convention here would be to prefix `last_non_direct_` in place of `session_` rather than suffix it. 

Last, non-direct is not session attribution, because it includes data from previous sessions so the `session_` prefix doesn't make sense. Secondly, it's easier to differentiate session source and last, non-direct session when they don't both begin `session_source*`.

models/staging/stg_ga4__sessions_traffic_sources.sql
dgitis, line 1: I'm not against moving commas to the start of the next line, but we really should have a style guide before we start shuffling things around that don't do anything.

dgitis, line 1: Wouldn't it make more sense to build `stg_ga4__sessions_traffic_sources_daily > dim_ga4__sessions_daily >  dim_ga4__sessions` and then delete this model entirely? That way the `dim_ga4__sessions` and the `dim_ga4__sessions_daily` models wouldn't both be hitting `base_ga4__events` when the build and it would make it so we don't need to update `stg_ga4__sessions_traffic_sources` and `stg_ga4__sessions_traffic_sources_daily` separately.

models/staging/stg_ga4__sessions_traffic_sources_last_non_direct_daily.sql
dgitis, line 1: I think adding and `attribution_window` var here makes more sense instead of hard-coding a 30-day lookback.

```
({{var('attribution_window',30) * 24 * 60 * 60}}  )
```

models/staging/stg_ga4__sessions_traffic_sources.sql
adamribaudo-velir: Yea good call. This was a byproduct of some enhancements I made to this model and then backed out (but left the comma changes)

We just acquired Brooklyn Data so I'd like to adopt their style guide moving forward: https://github.com/brooklyn-data/co/blob/main/sql_style_guide.md

adamribaudo-velir: Having `dim_ga4_sessions` build from `dim_ga4_sessions_daily` is definitely on the roadmap. I mentioned that in a comment a while back but I should create a specific issue for that. 

Updating `stg_ga4__sessions_traffic_sources` and `stg_ga4__sessions_traffic_sources_daily` separately is intentional as it allows users to access multi-day sessions (stg_ga4__sessions_traffic_sources) or single-day sessions (stg_ga4__sessions_traffic_sources_daily). I personally don't think the value of multi-day sessions is worth the complexity and cost, but I'm hesitant to remove the option entirely.

So I think both these concerns are outside the bounds of this PR, but definitely valid points.

models/staging/stg_ga4__sessions_traffic_sources_last_non_direct_daily.sql
adamribaudo-velir: I tried that but BigQuery complains. Apparently you can't put math operations in that clause. 

I could set the integer as a variable, though, and default to 30 days. I'll do that.

models/marts/core/dim_ga4__sessions_daily.yml
```diff
@@ -20,3 +20,7 @@ models:
       - name: session_partition_date
         description: >
           Date associated with the session_partition_key. Used to partition the table. Filter on this column to optimize query cost and performance. 
+      - name: session_source
+        description: The source of the session based on the events contained within this particular session. For last non-direct source, see session_source_last_non_direct
```
dgitis, line 24: Update the field reference to `last_non_direct_source`.

dgitis (approved):
There's a small documentation fix that's still needed. Otherwise, this looks good. There's no need to submit this to me for review again.

PR #200, opened 2023-05-16
update standard ecommerce params

dgitis:
## Description & motivation
I realize now that when I set up the first recommended events, I only included required fields as [documented here](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtm). In other cases, I or others included fields that aren't required in the recommended events, like `item_list_name` and `item_list_id` in the `view_item_list` event.

This PR adds those non-required parameters to ecommerce events.

I realize now that we should have a discussion on how to handle it. We either only add the required fields and then make our documentation clear that fields that aren't marked as required need to be added as custom parameters, or we default to adding all of the parameters listed in the [recommended events docs](https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtm).

## Checklist
- [] I have updated the README.md (if applicable)

### DIFF
--- a/models/staging/recommended_events/stg_ga4__event_add_to_cart.sql
+++ b/models/staging/recommended_events/stg_ga4__event_add_to_cart.sql
@@ -5,6 +5,8 @@
 }}
  with add_to_cart_with_params as (
    select *except (items),
+   {{ ga4.unnest_key('event_params', 'currency') }},
+   {{ ga4.unnest_key('event_params', 'value', 'double_value') }},
    (select items FROM UNNEST(items) items LIMIT 1) as items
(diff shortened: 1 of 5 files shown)


dgitis:
I should have opened an Issue to discuss this first, but only realized what the issue was when I was preparing a PR. @adamribaudo  and @willbryant, please comment on whether to include event parameters from recommended events that are documented but not required in our staging models.

adamribaudo-velir:
I don't have a strong opinion here. This is clearly an improvement from the current state so I'm happy to merge.

adamribaudo-velir (approved):

PR #201, opened 2023-05-20
Allow columns to be renamed regarding conversion events

YutaNakamura1413:
### Current Issue:
Build of `stg_ga4__page_conversions` model fails when the value set for `conversion_events` is a special character that does not correspond to a column name.

### Soluton:
Define a new optional variable `conversion_event_column_renamers` whose key matches `conversion_events` and assign the column name renamed at query generation time.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,17 @@ vars:
     conversion_events:['purchase','download']
 ```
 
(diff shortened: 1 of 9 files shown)


adamribaudo-velir:
Thanks, can you provide an example configuration and event name that breaks? I'd like to recreate the issue myself.

PR #228, opened 2023-06-01
fix code getting accidentally commented out

dgitis:
## Description & motivation

Something in how dbt processes macros in pre-hooks causes code without line breaks to appear on a single line in production but not development. This results in code that we want getting commented out when there isn't a blank line after the comment, but only in production.

This results in the following code running in production on dbt Cloud.

```
    -- If incremental, then use static_incremental_days variable to find earliest shard to copyCREATE OR REPLACE TABLE `project.dataset.events_intraday_20230530snip1` CLONE `project.dataset.events_intraday_20230530`;
```

This PR results in something like this code running.

```
    -- If incremental, then use static_incremental_days variable to find earliest shard to copy
            
            
            
                    CREATE OR REPLACE TABLE `project.dataset.events_intraday_20230530snip1` CLONE `project.dataset.events_intraday_20230530snip2`;
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Example:
 ```
 vars:
   ga4:
-    frequency:"daily+streaming"
+    frequency: "daily+streaming"
 ```
 
 # Connecting to BigQuery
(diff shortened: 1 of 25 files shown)


willbryant (approved):
WTAF, but thanks

adamribaudo-velir:
I can clarify. Using `{%-` will remove whitespace generated by the jinja script while `{%` will include the literal whitespace used inside the jinja script. 

Let me make that change, commit, and @dgitis please retest

adamribaudo-velir:
all set

dgitis:
I tried that but only removing the ticks from the jinja immediately surrounding the `clone` statements and it didn't do anything but this latest version with the changes that you made worked, albeit with a lot of whitepace between the comment and the clone statement.

dgitis:
I seem to have started branch behind by a few commits. This should be ready to merge now @adamribaudo-velir .

Issue #241, opened 2023-06-26
Package should accept a range of dbt_utils versions


Stephen986:
Specifying a specific version of dbt_utils in packages.yml makes it very hard for companies to use this package, as they may have other packages that require another version.  I already have one client who created their own version of this package to support dbt_utils 1.1 back when this package was on utils 1.0.

Instead, why not have the packages.yml for this package accept any version of utils between 1.0 and before 2.0?

PR #242, opened 2023-06-27
Allow range of dbt_utils versions

Stephen986:
## Description & motivation

PR for ticket: https://github.com/Velir/dbt-ga4/issues/241 .

Since this package is not using safe_subtract, it doesn't require dbt 1.1 (and certainly doesn't require 1.1.1 which is just a readme change).  On the other hand, future versions before 2.0 are not supposed to be breaking changes, so it should be safe to allow them.  Allowing a broad range of dbt_utils versions makes it much easier for companies to use this package along with other packages that also have dbt_utils version requirements.

## Checklist
- [ ] I have updated the README.md (if applicable) (N/A)
- [ ] I have added tests & descriptions to my models (and macros if applicable) (N/A)
- [X] I have run `dbt test` and `python -m pytest .` to validate existing tests

### DIFF
--- a/packages.yml
+++ b/packages.yml
@@ -1,3 +1,3 @@
 packages:
   - package: dbt-labs/dbt_utils
-    version: 1.1.1
+    version: [">=1.0.0", "<2.0.0"]


adamribaudo-velir:
thanks!

adamribaudo-velir (approved):

PR #245, opened 2023-06-30
Improvements to channel mapping based on Google's definitions

### LINKED ISSUE #237: Channel definition mis-match results in SMS traffic attributing as Organic Social
This rule in our `default_channel_grouping` macro

```
  when REGEXP_CONTAINS({{source}}, r"^(facebook|instagram|pinterest|reddit|twitter|linkedin)") = true
    or REGEXP_CONTAINS({{medium}}, r"^(social|social-network|social-media|sm|social network|social media)") = true
    or {{source_category}} = 'SOURCE_CATEGORY_SOCIAL' 
    then 'Organic Social'
```
precedes this rule.

```
  when {{medium}} = 'sms'
    then 'SMS'
```
The `regexp_contains(medium, r'sm')` rule matches sms and thus prevents the SMS rule from firing.

[Google's definition](https://support.google.com/analytics/answer/9756891?hl=en) is _Medium is one of (“social”, “social-network”, “social-media”, “sm”, “social network”, “social media”)_.

Is there a reason why our rule doesn't match here? Or was that just an oversight.

We can either fix this by switching the rule to `medium in (“social”, “social-network”, “social-media”, “sm”, “social network”, “social media”)` and match Google's definition or we can move the SMS rule to fire before the Organic Social rule.

Do you have any preferences @adamribaudo-velir ?

adamribaudo-velir:
## Description & motivation
Certain channel mappings were out of alignment with Google's definitions located here: https://support.google.com/analytics/answer/9756891?hl=en

This PR may not be exhaustive, but it is definitely an improvement over the prior mappings.

Also updated formatting to make it easier to read.

Closes #237 

## Checklist

### DIFF
--- a/macros/default_channel_grouping.sql
+++ b/macros/default_channel_grouping.sql
@@ -74,6 +74,7 @@ case
     then 'Audio'
   when 
     {{medium}} = 'sms'
+      or {{source}} = 'sms'
     then 'SMS'
   when 
     REGEXP_CONTAINS({{medium}}, r"(mobile|notification|push$)") or {{source}} = 'firebase'
--- a/macros/default_channel_grouping.sql
+++ b/macros/default_channel_grouping.sql
@@ -8,12 +8,16 @@
 
 {% macro default__default_channel_grouping(source, medium, source_category) %}
 case 
-  when {{source}} is null and {{medium}} is null 
+  when 
+    (
+      {{source}} is null 
+        and {{medium}} is null
+    )
+    or (
+      {{source}} = '(direct)'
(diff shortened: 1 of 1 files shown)


dgitis (approved):
This is all good. 

I noticed that some of the definitions changed on Google's support document since I originally reported this. I expect these will go out of sync again.

We also might want to think about requesting review before making style guideline changes as the style guideline changes make the change highlighting pretty much useless.

I was happy to do this myself. I just wanted to check that there wasn't a deliberate reason for having different definitions.

PR #252, opened 2023-07-27
Query param extraction variable (can be used to pull out click IDs)

### LINKED ISSUE #234: Configurable session-scope click ID fields from referring site

adamribaudo-velir:
## Description & motivation

Resolves #234 

This PR adds a new project-level configuration variable, `query_parameter_extraction` which accepts an array of values. Those values are converted into new columns within `stg_ga4__events`. 1 column for each parameter extracted.

This is useful for pulling out click IDs from the `page_location` query params and joining them with advertising data sets. 

## Checklist


dgitis (commented):
One of the points of discussion between @willbryant and me on the topic of click IDs was that advertising click IDs were **mutually exclusive**. You can't have clicks from multiple ad platforms at once.

The approach in this PR is not a waste but it results in much larger tables than necessary as `gclid` and `fbclid` never co-occur. The original approach of having a `click_type` and `click_value` solves most click ID requirements with just two columns.

`click_type = gclid` and `click_value = gclidvaluehere`
`click_type = fbclid` and `click_value = fbclidvaluehere`

It is possible that there are click parameters that aren't mutually exclusive, I'm thinking maybe cross-platform attribution plugins installed on your various ad platforms, but these aren't that common. 

I'm of the opinion that if someone is extracting a non-mutually-exclusive query parameter then they should just do that by overriding the `stg_ga4__events` model because it's unlikely that they will have more than 1 or 2 of these.

adamribaudo-velir:
@dgitis sure, a single click will result in mutually exclusive click IDs, but once the data team wants to join with this data wouldn't it be much easier to join the Facebook Ads data on a 'fbclick_param' column and join the Google Ads data on a 'gclid_param' column?  

A single solution for click IDs and any additional arbitrary query params seems a bit cleaner overall. BQ compresses at the column level and null values won't increase the size of the table.

dgitis:
I actually wasn't aware of BigQuery's handling of null values. That's better then I had feared.

dgitis (approved):

PR #256, opened 2023-07-28
fix combine property data to drop old intraday tables

### LINKED ISSUE #255: combine_property_data does not drop old intraday tables
When using the multi-property + streaming import feature introduced with #187, then the combined dataset accumulates intraday tables over time. As soon as for the intraday table `analytics_xxxxx.events_intraday_yyyyyyyy` its corresponding daily table `analytics_xxxxx.events_yyyyyyyy` becomes available, `combine_property_data` will copy it but leave the old `<combined_dataset>.events_intraday_yyyyyyyyxxxxx` behind.
This ultimately leads to duplicated data in `base_ga4__events` that is not sufficiently removed by the final `qualify row_number() .... = 1` statement.

FlorianASchroeder:
## Description & motivation
Resolves #255 

This PR drops old intraday tables in case their equivalent daily table becomes available and has been copied to the combined dataset.
It fixes the issue of:
* data duplication that is not sufficiently resolved by the deduplication in `base_ga4__events`
* waste of storage (and cost) of intraday tables clones

## Checklist

### DIFF
--- a/macros/combine_property_data.sql
+++ b/macros/combine_property_data.sql
@@ -17,7 +17,7 @@
 
     {% for property_id in var('property_ids') %}
         {%- set schema_name = "analytics_" + property_id|string -%}
-            {# Copy daily tables #}
+            {# Copy intraday tables #}
             {%- set relations = dbt_utils.get_relations_by_pattern(schema_pattern=schema_name, table_pattern='events_intraday_%', database=var('project')) -%}
             {% for relation in relations %}
                 {%- set relation_suffix = relation.identifier|replace('events_intraday_', '') -%}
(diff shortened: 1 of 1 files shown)


dgitis:
Thanks for reporting this and submitting a fix. I'll try creating a dummy dataset to test this fix on over the weekend.

dgitis (approved):
I duplicated a partition on a test project that runs only intraday, named it so that it would get picked up as a daily table, updated my test project to process the duplicate project twice by assigning the project ID two times, and then ran this PR against it. It looks like it works.

There's no intraday on that particular date. Thanks for submitting this fix.

adamribaudo-velir:
merging. thanks!

PR #250, opened 2023-07-19
check for intraday table before processing union operation

adamribaudo-velir:
## Description & motivation
Resolves #249

When implementing #238, a bug was introduced whereby the base events model would attempt to union intraday data even if it did not exist.

This PR uses the dbt.utils `dbt_utils.get_relations_by_pattern` function to evaluate whether the intraday table exists before attempting to union its data. It only does this check when NOT combining data from multiple properties (the combine_property_data macro has its own check to see if the intraday table exists)

## Checklist
- [x] I have verified that these changes work locally
- [na] I have updated the README.md (if applicable)
- [na] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have run `dbt test` and `python -m pytest .` to validate existing tests

PR #257, opened 2023-07-30
Fix streaming support for multi property configurations

### LINKED ISSUE #254: Multi-Property Support with streaming intraday tables
Due to a recent change in #250 it is not possible anymore to use multiple properties together with streaming support.
This breaks the feature introduced by #187.

Can this be enabled again anytime soon?

adamribaudo-velir:
## Description & motivation
Resolves #254 

As noted by @FlorianASchroeder , PR #250 allowed the union of intraday data only for single-property configurations. 

This update removes that limitation by leveraging the fact that both daily and intraday event tables start with 'events_'. This means we can use a wildcard operation to query `events_*` to pull both daily and intraday records. Duplicate records will be deduped as part of the `base_ga4__events` qualify statement. 

As a result, we can remove the intraday source from `src_ga4.yml`

## Checklist

### DIFF
--- a/models/staging/base/base_ga4__events.sql
+++ b/models/staging/base/base_ga4__events.sql
@@ -2,7 +2,9 @@
 {% for i in range(var('static_incremental_days')) %}
     {% set partitions_to_replace = partitions_to_replace.append('date_sub(current_date, interval ' + (i+1)|string + ' day)') %}
 {% endfor %}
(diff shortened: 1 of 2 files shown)


dgitis (approved):
My concern with this PR is that the `qualify` statement at the end won't deduplicate in all situations. I am not aware of any `event_params` that get added when Google processes the GA4 data, but just one field would cause the deduplication to stop working.

I'm also concerned that deduplicating this way will take a lot of unnecessary processing, but I can't think of a better way of doing this.

PR #265, opened 2023-08-15
Set "Direct" instead "(none)" as default value

clemens7haar:
## Description & motivation

The session default channel grouping column contained "(none)" instead of "Direct" if no other value was found.

Some time ago the change was already merged (#157). Probably it was overwritten by mistake. That is why I am resubmitting it.

## Checklist


adamribaudo-velir (approved):
Strange. Not sure how this reverted. Thanks!

PR #272, opened 2023-09-26
Add adapter dispatch pattern for path extraction

adamribaudo-velir:
## Description & motivation
It can be helpful for users of the package to override the page path extraction to accommodate implementation issues. This PR applies the adapter dispatch pattern to the `extract_page_path` macro.

## Checklist


adamribaudo-velir:
There is an issue with the unit tests that we're working through. I'm going to merge this in the meantime.

PR #277, opened 2023-10-17
Dedup transactions

### LINKED ISSUE #193: Deduplicating transactions
GA4 by default deduplicates purchases with the same `transaction_id`. We should probably do the same.

I'm currently chasing down a bunch of duplicate transactions and I'm noticing a mix of implementation issues, users reloading the confirmation page, and bots that seem to want to crawl these pages several months after the fact.

I think, for performance reasons, we should just commit to deduplicating transactions within the period being processed and not look back beyond that.

We could do this in `stg_ga4__event_purchase`, but we would need to unnest the `transaction_id` from the items array to do it there. Another option is to create a sample `fct_ga4__event_purchase` table that people can enable and fix it there. 

We may also want to consider removing the duplicates from the items table.

Additionally, purchase events where `transaction_id` is null should not be tracked as these are invalid according to Google, but they come through to the BigQuery export and currently get processed by our package.

Regardless of what we choose, I think we want to keep the purchase with the oldest timestamp.

I think I do the most ecommerce work, but I'd like @adamribaudo and @willbryant to weigh in here. Adam, in particular, has preferred to keep our event staging models consistent and one of the options here involves quite a few changes to one of those models.

dgitis:
## Description & motivation
Resolves #193.

This PR deduplicates transactions on `transaction_id`  within  the current processing window as determined by the `static_incremental_days` variable.

There are a number of possible issues could be fixed here but many of them are in conflict with one another so I chose to keep this as simple as possible.

## Checklist

### DIFF
--- a/models/staging/recommended_events/README.md
+++ b/models/staging/recommended_events/README.md
@@ -28,3 +28,13 @@ models:
 ```
 
 Not all recommended events have been implemented. If you need a specific event, please consider creating a pull request with the model that you need in the [dbt-ga4 GitHub repository](https://github.com/Velir/dbt-ga4).
+
+## Purchase Event Transaction Deduplication
+
+The `stg_ga4__event_purchase_deduplicated` model builds on the `sgt_ga4__event_purchase` model. It is disabled by default and thus needs to be enabled along with the `stg_ga4__event_purchase` model.
+
+The model only processes purchase events that fall within the window as defined by `static_incremental_days` and can only reliably be expected to deduplicate purchase events occurring in the same day.
+
+The model provides a highly-performant, minimum-viable product for this feature returning only data from the first purchase event with a matching `transaction_id` within the processing window.
+
+You are encouraged to copy this model to your project and customize it there should this MVP be insufficient for your needs.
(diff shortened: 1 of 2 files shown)


dgitis:
Just for reference, the performance additions in [52868e5](https://github.com/Velir/dbt-ga4/pull/277/commits/52868e5d1449aab52608487effdb4c1c7486103a) process 136.46 MB on a client test site that's been running for over a year compared with 8.8 GB from before the optimization.

The changes reduce the effectiveness when deduplicating, but I prefer to default to the performant option and let people undo that as needed.

adamribaudo-velir:
The business logic seems sound, but we could use the QUALIFY statement to cut the amount of SQL, remove the 'distinct', and remove a CTE. See here for example: https://github.com/Velir/dbt-ga4/blob/f2f03a2475b3e567a452510e04e01a3464ca59c2/models/staging/base/base_ga4__events.sql#L36C1-L36C26

Something like this I think:
```
qualify row_number() over(
 partition by transaction_id 
        order by 
            event_timestamp asc rows between unbounded preceding
            and unbounded following
) = 1
```

dgitis:
Review the new version when you get an chance. It seems to work fine on my test site. 

The processing cost on both versions is equal.

adamribaudo-velir (approved):

Issue #273, opened 2023-09-28
Default channel grouping not in line with GA4 default channel  grouping


erikverheij:
Hello,

Firstly, I'd like to express my gratitude for this incredible package; it has been immensely beneficial in my work.

While reviewing default_channel_grouping.sql, I noticed that the "Paid Other" channel group, which is a default in GA4, appears to be missing. The documentation from Google Analytics for the "Paid Other" channel group is provided here: [GA4 Default Channel Groups](https://support.google.com/analytics/answer/9756891).

For reference, the "Paid Other" criteria is defined as:
Medium matches regex `^(.*cp.*\|ppc\|retargeting\|paid.*)$`

Upon further inspection, and while comparing with the GA4 documentation, I observed a few discrepancies:

- **Other Advertising:** It appears this might not be a part of GA4's default channels anymore.
- **Push Notifications:** The GA4 documentation now refers to this as Mobile Push Notifications.
- **Cross-network:** This is defined in GA4's default channels, but I couldn't locate its presence in macros/default_channel_grouping.sql.

For visual reference, please see the image provided:

![image](https://github.com/Velir/dbt-ga4/assets/3001961/ef228f20-de37-48fd-aa5f-e041bdc3fa80)

I hope these observations prove useful, and I'm eager to contribute or assist further if needed. Thank you for your continuous effort in maintaining and improving this repository.

Kind regards, Erik

adamribaudo-velir:
Very useful! Thanks so much. Do you feel comfortable submitting a PR to resolve these discrepancies? 

It does make me wonder if Google changes these from time to time as I thought our implementation was correct as of ~6 months ago 🤔

erikverheij:
Thanks for your reply :) I'll create a PR in the upcoming weeks!

PR #282, opened 2023-11-03
Align default channel groupings with GA4 docs

erikverheij:
## Description & motivation
Align default channel groupings with GA4 docs. The following things have been changed:

* Order of statements aligned with order in GA4 docs.
* Refactored test: Included more test cases and combined input with the expected channel for easier readability.
* Added campaign param to default_channel_grouping() method as it was needed for some mappings.
* Included GA4 descriptions to make it easier to maintain.
* Updated ga4_source_categories.csv based on PDF download from https://support.google.com/analytics/answer/9756891?hl=en#list
* Removed the reference to https://github.com/llooker/ga_four_block_dev as I don't think their implementation is used anymore.

Also see issue #273.

## Checklist
- [ ✅] I have verified that these changes work locally
- [ not applicable] I have updated the README.md (if applicable)
- [✅] I have added tests & descriptions to my models (and macros if applicable)
- [✅ ] I have run `dbt test` and `python -m pytest .` to validate existing tests


adamribaudo-velir (approved):
Looks good! Thanks for the thorough unit test. Just had to replace curly quotes as it was causing a text parsing issue for me.

erikverheij:
Great! Thanks for reviewing & merging it so quickly!

PR #283, opened 2023-11-10
Explicit Item and Ecommerce field selection

### LINKED ISSUE #281: The new items format for custom parameters causes base models to error
Here's the error that I'm seeing.

```
12:55:50  Completed with 1 error and 0 warnings:
12:55:50  Database Error in model base_ga4__events (models/staging/ga4/base/base_ga4__events.sql)
12:55:50    Value has type ARRAY> which cannot be inserted into column items, which has type ARRAY> at [171:1194]
```

And here's the new schema.

image

I first noticed this on a client site that hasn't been touched in a few weeks so this is almost certainly the cause of the error.

dgitis:
## Description & motivation
In order to prevent Google changes to the `items` and `ecommerce` records breaking the package, this PR hard-codes the fields output by the `base_select` macro which will prevent changes from breaking and allow users to override the macro and incorporate changes without waiting on a package update.

Resolves #281 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,13 @@ packages:
 packages:
   - local: ../dbt-ga4
 ```
+## Version Upgrade
+Since the release of version `5.0.0`, Google changed format of the `items` record which caused the `base_ga4__events` model to fail. This was fixed with a simple full refresh, but that cost extra processing.
+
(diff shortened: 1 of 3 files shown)


dgitis:
The ReadMe changes may not be necessary. I thought I'd found my test project which has data going back a long had a couple of fields missing, but it was an error on my part compounded by my needing to run a full refresh on my test environment which stopped me from catching the missing fields.

adamribaudo-velir (approved):

PR #294, opened 2023-12-29
Allow distinct source and target projects

adamribaudo-velir:
## Description & motivation
This update allows users to set distinct source and target projects. This helps in scenarios where the dbt user is not able to write to the source project. The target project is pulled from the current profile (`target.project`).

!!Certain variables have been renamed which introduce breaking changes!!

`dataset` and `project` no longer exist as the `target.dataset` and `target.project` will be used.
New variable: `source_project` is now required to indicate the source of the GA4 data
New variable: `combined_dataset` is required when sourcing from multiple properties. This specifies where to copy the cloned data. The `target.project` will be used to clone data.
`property_ids` variable is an array and is used to indicate which property IDs to source from. Can be used for single or multi-property instances (ex: [1111111] for 1 property or [1111111,2222222] for 2, etc)

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,25 @@ This package assumes that you have an existing DBT project with a BigQuery profi
 ```
 vars:
   ga4:
-    project: "your_gcp_project"
-    dataset: "your_ga4_dataset"
+    source_project: "my_source_gcp_project" # Project that contains raw GA4 data
+    property_ids: [11111111] # Array of properties to process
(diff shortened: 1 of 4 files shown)


adamribaudo-velir:
@hugocaillol

dgitis (approved):

hugocaillol:
@adamribaudo-velir 
Admirable reactivity <3

PR #295, opened 2024-01-05
fix fct_ga4__pages model and add tests to catch similar issues

### LINKED ISSUE #274: fct_ga4__pages duplicate dimension fields
It seems like the `fct_ga4__pages` table is creating duplicate rows where it should now and that removing the `page_engagement_key` fields fixes the problem.

From Tony Tushar on dbt Slack

> So if I’m seeing duplicates across all dimension fields in that table, but seeing differences on say a single metric/fact column, is it wrong to group by date further in that table, as you’re confirming it is not by hour anymore.

--- snip ---

> Lastly, will mention that seems like the removal of page_engagement_key in the page_engagement CTE for fct_ga4__pages is causing duplicates.

dgitis:
## Description & motivation

The addition of `page_engagement_key` which is built from the `page_referrer`  where page_view events are concerned was causing multiple rows with the same `page_location` and `event_date_dt` in the first CTE.

This PR adds a second aggregation in the second CTE to eliminate this redundancy and adds tests to the model to catch these issues in the future.

The `fct_ga4__pages` model will need a full refresh when upgrading to a version containing these changes as the new tests will trigger an error.

Resolves #274 

## Checklist

### DIFF
--- a/models/marts/core/core.yml
+++ b/models/marts/core/core.yml
@@ -14,13 +14,6 @@ models:
         description: Hashed combination of user_pseudo_id and stream_id
         tests:
           - unique
-  - name: fct_ga4__pages
-    description: Incremental model with page metrics such as visits, users, new_users, entrances and exits as well as configurable conversion counts. 
-    columns:
-      - name: total_engagement_time_msec
-        description: The total engagement time for that page_location.
-      - name: avg_engagement_time_denominator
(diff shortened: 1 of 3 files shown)


adamribaudo-velir:
Looks like the unique test can fail when there is a page with the same page location but different page titles (not uncommon for translated content)

dgitis:
This latest version removes `page_path` and `page_title` from the `fct_ga4__pages` model which will help this model pass its uniqueness test.

adamribaudo-velir (approved):
tests passed this time!

PR #304, opened 2024-02-27
Fix to filter based on dates equal to or later than the start_date

yamotech:
## Description & motivation

In the original code `cast(replace(_table_suffix, 'intraday_', '') as int64)`, larger values than the start_date were being retrieved, resulting in unsuccessful filtering.

```sql
select distinct
    _table_suffix
    , cast(replace(_table_suffix, 'intraday_', '') as int64) as original
    , cast(left(replace(_table_suffix, 'intraday_', ''), 8) as int64) as correct
from `{{ target.project }}`.`{{ var('combined_dataset') }}`.`events_*`
limit 1
```

- original: `{{ YYYYMMDD }}{{ property_id }}`
- correct: `{{ YYYYMMDD }}`

## Checklist

### DIFF
--- a/models/staging/base/base_ga4__events.sql
+++ b/models/staging/base/base_ga4__events.sql
@@ -21,7 +21,7 @@ with source as (
     select
         {{ ga4.base_select_source() }}
     from {{ source('ga4', 'events') }}
-    where cast(replace(_table_suffix, 'intraday_', '') as int64) >= {{var('start_date')}}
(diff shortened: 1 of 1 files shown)


adamribaudo-velir:
Good catch!

adamribaudo-velir (approved):

adamribaudo-velir:
I can do that tomorrow. You can point your packages.yml to this specific commit in the meantime if you need it sooner.

PR #305, opened 2024-03-05
Fix uniqueness test for fct_ga4__pages

yamotech:
## Description & motivation

Fix uniqueness test for fct_ga4__pages.

Multiple records can have the same event_date_dt and page_location.
Adding stream_id as a condition makes them unique.

> Data stream: Lives within a property, and is the source of data from an app or website. The best practice is to use a maximum of 3 data streams per property: 1 single web data stream to measure the web user journey and 1 app data stream each for iOS and Android.
>
> [[GA4] Google Analytics account structure - Analytics Help](https://support.google.com/analytics/answer/9679158?hl=en)

## Checklist


dgitis (approved):
Good catch, thank you.

PR #306, opened 2024-03-12
fix cpc classification and add test

### LINKED ISSUE #302: GA4 raw data format can break google / organic gclid fix
It seems that Google has started sometimes setting `source` and `medium` values to `google / organic` when a visitor clicks an ad.

This breaks the code that we use to fix `google / cpc` source/medium values that are mis-attributed to `google / organic` in the `stg_ga4__events` file.

```
detect_gclid as (
    select
        * except (event_source, event_medium, event_campaign),
        case
            when (page_location like '%gclid%' and event_source is null) then "google"
            else event_source
        end as event_source,
        case
            when (page_location like '%gclid%' and event_medium is null) then "cpc"
            else event_medium
        end as event_medium,
        case
            when (page_location like '%gclid%' and event_campaign is null) then "(cpc)"
            else event_campaign
        end as event_campaign
    from include_event_key
```

Here's a graph of raw event counts by day with a `page_location` containing 'gclid' by `medium` for one site:

image

This image shows `source` and `medium` appearing in the `event_params.key` fields. The values are truncated to protect private data, but the values for source / medium are 'google / organic'.

image

I've confirmed that this is an issue on older GA4 properties.

I know we don't like adding variables whenever we can avoid it, but I think the fix has to be that we add a variable like `use_google_ads_manual_utm_for_attribution` and if that's true then we use `event_source`, `event_medium` etc... in `stg_ga4__events` and otherwise default to automatically setting source / medium when a gclid is detected.

We also need to verify that manually tagged UTMs from Google Ads set the source / medium to the tagged values rather than google / organic.

What are your thoughts? Are any of you working on a site that manually tags UTMs from Google Ads?

@adamribaudo @willbryant

dgitis:
## Description & motivation
Resolves #302

Google has started sometimes setting source and medium values to google / organic when a visitor clicks an ad.

This breaks the code that we use to fix google / cpc source/medium values that are mis-attributed to google / organic in the stg_ga4__events file.

This PR fixes that issue and adds a test to detect when it might happens again.

## Checklist
- [ y] I have added tests & descriptions to my models (and macros if applicable)

### DIFF
--- a/models/staging/stg_ga4__events.sql
+++ b/models/staging/stg_ga4__events.sql
@@ -38,10 +38,12 @@ detect_gclid as (
         end as event_source,
         case
             when (page_location like '%gclid%' and event_medium is null) then "cpc"
+            when (page_location like '%gclid%' and event_medium = 'organic') then "cpc"
             else event_medium
         end as event_medium,
         case
             when (page_location like '%gclid%' and event_campaign is null) then "(cpc)"
+            when (page_location like '%gclid%' and event_campaign = 'organic') then "(cpc)"
             else event_campaign
         end as event_campaign
     from include_event_key
--- a/tests/page_location_with_gclid_is_cpc.sql
+++ b/tests/page_location_with_gclid_is_cpc.sql
@@ -1,12 +1,6 @@
 -- Google has changed the combination of parameters that are used to identify a CPC source in the past.
 -- In order to detect new changes, this test checks that a page_location with a gclid is classified as cpc.
--- The purpose of this test is to detect changes in the source classification of CPC traffic.
--- As a result, it defaults to testing only the last static_incremental_days worth of tests.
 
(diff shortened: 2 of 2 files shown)


dgitis:
@cfrye2 I've just published a branch with your fix and added a test.

Do you still have a client that's getting cpc assigned to organic?

If so, can you copy [this file](https://github.com/Velir/dbt-ga4/blob/gclid/tests/page_location_with_gclid_is_cpc.sql) into  your tests folder and run `dbt test --select page_location_with_gclid_is_cpc`?

I haven't been able to verify that this test fails when it should.

dgitis:
I just confirmed with @cfrye2 that the test with `original_page_location` now fails when it is supposed to.

adamribaudo-velir (approved):

PR #316, opened 2024-03-28
feat/incremental-models

chienazazaz:
## Description & motivation

Convert some models materialization from table/view to incremental to reduce the amount of data processed at each update

## Checklist


adamribaudo-velir:
It looks like you accidentally committed your project settings in https://github.com/chienazazaz/dbt-ga4/blob/main/dbt_project.yml

Also, the goal of the PR seems to be to make the sessions models incremental, which is what the "_daily" models are for: https://github.com/chienazazaz/dbt-ga4/blob/main/models/marts/core/dim_ga4__sessions_daily.sql

GA4 sessions by default can last >1 day which is why the main sessions tables are not partitioned on day, but the `_daily` models are.

I'm going to close the PR given that its purpose is unclear and it's only barely documented. Please open an issue to discuss what you'd like the contribute before opening another PR.

Issue #285, opened 2023-11-10
Partition user and client tables


dgitis:
It would be good to have daily partitioned versions of the `dim_ga4__client_keys`, `fct_ga4__client_keys`, and `fct_ga4__user_ids` similar to what we have with sessions so that larger sites can disable the non-partitioned models without needing to customize.

The new GA4 user export tables are day partitioned.

I believe this should be related to #251 with us adding an optional cutoff date for when to start using Google's user export (because even when enabled, they didn't immediately start receiving all of the data) and merge the two sources of data in the daily tables and then build the non-day partitioned tables from the merged daily tables.

When comparing our client_key fields with the equivalent pseudonymous_users table in the new export, I think it is best that we set up our daily tables to contain basically the same data as is in the new export renamed and unnested to our usually standard. We then try to build as much as possible from before the cutoff into that table.

For the non-partitioned tables, do we try to maintain compatibility with our existing fields? For example, the `first_device_*` and `first_geo_*` fields don't have equivalents in the GA4 export.

While it would be nice to maintain compatibility, I personally don't use most of those fields. 

If others use them, then I'm happy to rebuild that downstream of the daily models. 

I am resistant to rebuilding that data on the daily models because if you're trying to reduce the costs by using just the daily models despite less accurate data then you probably won't want to do the look-ups required to enhance the daily models either. Particularly if you don't use the fields all that often.

Thoughts @adamribaudo @willbryant ?

adamribaudo-velir:
Waiting for access to a dataset that actually holds this data before weighing in. Should be soon.

PR #317, opened 2024-04-13
User Export

dgitis:
## Description & motivation
Work-in-progress resolving #285.

The goal of this PR is to support the `pseudonymous_user_id` and `user_id` tables.

My initial thought is that we should keep both the package's current user tables and new ones derived from the new user export options in GA4. The reason for this is to support both old and new installations.

README modifications are not done, but I expect we will want to add a new section on disabling user models that explains the differences between our various models and how to use them.

What defaults should we have for `+enable` on these tables?

This PR supports `audiences` fields from the export. While working with a test site that has implemented audiences, the data doesn't seem to be very useful unless its ID field joins with Google Ads audience exports. Despite this, I think we should leave this in because it needs to be enabled by a variable.

The `user_properties` fields were done without sample data. Naming and logic will likely need to be updated.

Multi-site is not currently supported.

We also might want to move the logic in `base_ga4___*` models to `stg_ga4__*` models.

I also need to review package naming conventions to ensure data, like geo data, is consistent with elsewhere in the package.

To-do:
- data marts
- review package naming conventions to ensure consistency
- `user_property` logic
- readme
- multi-site

Here are the [docs for the source tables](https://support.google.com/analytics/answer/12769371?hl=en&ref_topic=9359001&sjid=5665659258369477-NC).

## Checklist

### DIFF
--- a/macros/base_select.sql
+++ b/macros/base_select.sql
@@ -160,3 +160,48 @@
         ELSE 0
     END AS is_purchase
 {% endmacro %}
+
+{% macro base_select_usr_source() %}
+    {{ return(adapter.dispatch('base_select_usr_source', 'ga4')()) }}
+{% endmacro %}
+
+{% macro default__base_select_usr_source() %}
+    , user_info.last_active_timestamp_micros 
+    , user_info.user_first_touch_timestamp_micros 
+    , user_info.first_purchase_date 
+    , device.operating_system 
+    , device.category 
+    , device.mobile_brand_name 
+    , device.mobile_model_name 
+    , device.unified_screen_name 
+    , geo.city 
+    , geo.country 
+    , geo.continent 
+    , geo.region 
+    , user_ltv.revenue_in_usd 
+    , user_ltv.sessions 
+    , user_ltv.engagement_time_millis 
+    , user_ltv.purchases 
+    , user_ltv.engaged_sessions 
+    , user_ltv.session_duration_micros 
+    , predictions.in_app_purchase_score_7d 
+    , predictions.purchase_score_7d 
+    , predictions.churn_score_7d 
+    , predictions.revenue_28d_in_usd 
+    , privacy_info.is_limited_ad_tracking 
+    , privacy_info.is_ads_personalization_allowed 
+    , occurrence_date 
+    , last_updated_date
+    {% for up in var('user_properties', []) %} -- don't have sample data; need to verify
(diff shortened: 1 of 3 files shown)


dgitis:
@adamribaudo-velir, while I've verified that these work locally, I've only verified with a minimal amount of data. Multi-site changes were verified by setting the same property ID several times under `property_ids`. Obviously this is not ideal.

It would be best to verify this using a proper multi-site installation. I don't have any current clients, but I can ask some past clients.

I did not write any tests. [dbt recommends testing certain](https://docs.getdbt.com/docs/build/unit-tests#when-to-add-a-unit-test-to-your-model) things. The only logic worth testing , the multi-site macro, is not testable by the new framework.

I defaulted these new tables to `+enable: false`. 

I also ran `base_ga4__pseudonymous_users` in to `stg_ga4__client_ids`. I renamed at the staging model because that's where the client IDs are generated.

macros/combine_property_data.sql
```diff
@@ -6,41 +6,69 @@
     {% if not should_full_refresh() %}
         {# If incremental, then use static_incremental_days variable to find earliest shard to copy #}
         {%- set earliest_shard_to_retrieve = (modules.datetime.date.today() - modules.datetime.timedelta(days=var('static_incremental_days')))|string|replace("-", "")|int -%}
```
adamribaudo-velir, line 8: @dgitis I think you've introduced some duplicate code in this macro?

adamribaudo-velir (changes requested):

adamribaudo-velir:
I'm on board with the approach, but unable to run it due to this error. Couldn't find an obvious issue.  I can run the main branch fine with the same project configuration. 

```
20:08:18  Completed with 1 error and 0 warnings:
20:08:18
20:08:18    Compilation Error in model base_ga4__events (models\staging\base\base_ga4__events.sql)
  can only concatenate str (not "int") to str

  > in macro combine_property_data (macros\combine_property_data.sql)
  > called by macro run_hooks (macros\materializations\hooks.sql)
  > called by macro materialization_incremental_bigquery (macros\materializations\incremental.sql)
  > called by model base_ga4__events (models\staging\base\base_ga4__events.sql)
```

dgitis:
I fixed the duplicate code issues which seems to have fixed the error, not sure why though, and also updated a test that worked on my very small test site but wouldn't work on a larger site.

PR #331, opened 2024-07-03
Support for Custom Item Parameters

adamribaudo-velir:
## Description & motivation
This PR provides support for custom item parameters using a project variable of `custom_item_parameters` Ex:

```
vars:
  ga4:
    custom_item_parameters:
      - name: "item_variant_color"
        value_type: "string_value"
        rename_to: "color"
```

custom item parameters are exposed in the `stg_ga4__event_items` model, only. 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,18 @@ vars:
       - name: "country_code"
         value_type: "int_value"
 ```
+### Custom Item Parameters
+
+If you are capturing custom parameters at the item level (within items.item_params), you can specify these parameters and make them available in the `stg_ga4__event_items` model. This uses a similar syntax as the "Custom Parameters" feature described above:
+
+```
+vars:
+  ga4:
+    custom_item_parameters:
+      - name: "item_variant_color"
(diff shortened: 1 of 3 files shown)


dgitis (approved):
I don't have a test property that uses custom item properties, but the code looks good.

geo909:
Hey, thanks for the quick reply! 

I will be completely honest, I got confused, thought this PR was for default event-granularity parameters, not for item parameters. I'm quite new to ga4 so I'm still being confused with basic concepts, apologies. :(  I thought that it would make sense to add any _default_custom_parameters_ in the `stg_ga__events` model instead of only the `stg_ga4__event_<each_event>` model. But I don't want to hijack this PR, apologies. Let me know if there is any other issue about this or if it is a good idea to open one.

Regarding this PR, unfortunately we also don't have custom item parameters in our data so I cannot test. Running the code with mock custom_item_parameters, however, did create the corresponding fields in `stg_event_items`.

PR #338, opened 2024-08-06
Query string all

dgitis:
## Description & motivation
Initially, I wanted to add an `*all*` flag to the remove query string feature.

When writing the tests, I discovered that the remove query string only removed question and ampersand query strings but not hash strings which may not technically be query strings but are sometimes used as such. 

In fixing the regular expression, I ended up simplifying it and added tests to ensure that the new regex works as expected.

We may want to add the following to the release notes when pushing this live.

> The `query_parameter_exclusions` variable now removes fragments, hash `#` parameters, from `page_location` and `page_referrer`. In most cases this will be an improvement in functionality. However, if you have query parameter keys using question mark, `?`, or ampersand, `&`, that you want to remove and fragments that you don't want to remove, then you will need to override the `remove_query_parameters` macro.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,14 @@ vars:
     query_parameter_exclusions: ["gclid","fbclid","_ga"] 
 ```
 
+You can remove all query parameters by setting `query_parameter_exclusions` to `*all*`.
+
+```
+vars:
+  ga4:
+    query_parameter_exclusions: ["*all*"]
+```
+
 ### Query Parameter Extraction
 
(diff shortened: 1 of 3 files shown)


adamribaudo-velir:
Looks great. thanks for adding the unit test. I need to figure out how to get the dbt unit tests to run on PR updates, but I can do that separately. The unit tests passed.

adamribaudo-velir (approved):

Issue #340, opened 2024-08-15
Handling null ga_session_id in event_key creation


EvavW:
When `ga_session_id` is null, so is `event_key`. This is problematic when using `event_key` to count unique events. 

Null `event_key`s could be avoided by using `array_to_string` to join keys rather than `concat`. such as:
```
include_session_key as (
    select 
        *,
        to_base64(md5(array_to_string([client_key, cast(session_id as string)], ""))) as session_key
    from include_client_key
),
```

This should generate the same keys as the current approach when `ga_session_id` is not null. This same approach could also handle null `user_pseudo_id`s.
Are there any concerns with this approach?

dgitis:
At first glance, I'm on board with this change.

We would also need to filter out the formerly null session keys from the down-stream session tables so that metrics don't shift suddenly or add a configuration that defaults to filtering these out. 

For consistencies sake, we probably want to filter out events with the new null event keys as well and add a default configuration here too.

Longer term, we should think about adding modeling options or instructions to the package similar to GA4's data modeling that would use these "null" key events and sessions.

I would imagine that the `fct_ga4__pages` model would work well as a sample, how-to-model missing data, summing metrics from `stg_ga4__event_page_view` and `stg_ga4__event_page_view_modeled` by `page_location`.

I don't actually know how to do the actual modeling? Is there a statistical function in BQ that does this already?

We don't actually need to show modeling in the package but I do think that the package should give people the option to model their data and that means that we need to handle this null issue with that in mind.

@adamribaudo-velir do you have any thoughts on handling null `ga_session_id`.

dgitis:
Actually, it occurs to me that the modeled metrics would be a simple calculation of figuring out the metric value per event or per session, multiplying that value by the number of non consented events or sessions and then adding that to the actual metric: `(sum(consented_metric_value)/count(consented_rows)) * count(non_consented_rows) + sum(consented_metric_value)`. 

This makes sense to do as a macro. 

Also, instead of having two staging models, like in my last comment, we include the consented and non-consented rows in the same model and then filter out the non-consented in our marts unless someone configures the filter variable to include non-consented data.

I see this as being a two stage change. 

In stage 1, we implement @EvavW's suggestion and add a `process_non_consented_data` variable that defaults to false and then conditionally include or filter these non-consented values at the end of `stg_ga4__events`.

[...]

PR #341, opened 2024-08-22
Issue #340 fill null event key

EvavW:
## Description & motivation
Follows discussion in Issue #340 
### Changes:
* Use array_to_string rather than concat to create event_key
* replace session_key with client_key and session_id in event_key creation
* Add warning for users that this could result in duplicate event_keys in cases where distinction between events depends on one or both of those values being not null.

### Rationale:
Generate event_key even when session_key is null (either because of null user_psuedo_id or null ga_session_id). This will improve the usability of event_key as a tool to count unique events. 

Use both client_key and session_id in key creation (instead of session_key) so that a null value in either user_pseudo_id or ga_session_id will not nullify the other. This prevents duplicate event_keys. For example:
- In a case where user_pseudo_id is available but ga_session_id is null. This results in a null session_key. If that session_key is used to create event_key for two events for which all other inputs to key generation are identical then there will be duplicates. Using both client_key and session_id in event_key generation prevents this.
- Of course in the case that both user_pseudo_id and ga_session_id are null and all other inputs to key generation are identical there will still be duplicates.

## Checklist

### DIFF
--- a/models/staging/stg_ga4__events.sql
+++ b/models/staging/stg_ga4__events.sql
@@ -22,11 +22,16 @@ include_session_partition_key as (
         CONCAT(session_key, CAST(event_date_dt as STRING)) as session_partition_key
     from include_session_key
 ),
--- Add unique key for events
+-- Add unique key for events. Potential to not be unique if session_key is null and uniqueness depends on differentiation by that value.
 include_event_key as (
     select 
         *,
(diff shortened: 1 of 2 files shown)


adamribaudo-velir (approved):
Looks good. Well-documented as well. Hope to see you contribute more!

EvavW:
Thanks for the review @adamribaudo-velir! Do you have an idea of when this can be merged and released?

adamribaudo-velir:
Released https://github.com/Velir/dbt-ga4/releases/tag/6.1.0

PR #347, opened 2024-10-27
Add collected_traffic_source record and nested columns

adamribaudo-velir:
## Description & motivation

Adding the columns included in the `collected_traffic_source` RECORD column. These should match 1:1 with the equivalent parameters pulled out of the `event_params` column. They are included for completeness as Google now breaks them out of the `event_params` column and into `collected_traffic_source`

## Checklist
- [NA] I have updated the README.md (if applicable)
- [NA] I have added tests & descriptions to my models (and macros if applicable)
- [X] I have run `dbt test` and `python -m pytest .` to validate existing tests

### DIFF
--- a/macros/base_select.sql
+++ b/macros/base_select.sql
@@ -35,6 +35,7 @@
     , ecommerce.unique_items
     , ecommerce.transaction_id
     , items
+    , collected_traffic_source
     , {%- if  var('combined_dataset', false) != false %} cast(left(regexp_replace(_table_suffix, r'^(intraday_)?\d{8}', ''), 100) as int64)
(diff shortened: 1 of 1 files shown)


dgitis:
What do you think about allowing people to enable these fields via a variable? 

This change will cause the base model to error when people upgrade and it doesn't really provide new features since we already calculate the same data. By making it configurable, we can make it so people can upgrade without erroring. 

Is this part of a longer plan to simplify the attribution models?

dgitis (approved):

dgitis:
I thought that appending fields was an optional configuration, but it appears that I was wrong.

PR #355, opened 2024-12-26
User single site

dgitis:
## Description & motivation
GA4 user export tables for single-site dbt-GA4. 

This PR adds base and staging models for the native user export. The main feature of these tables is the ability to unnest user properties and audiences using variables. They are disabled by default but the Readme covers the range of user tables.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,47 +173,6 @@ vars:
         value_type: "int_value"
 ```
 
-### User Properties
-
-User properties are provided by GA4 in the `user_properties` repeated field. The most recent user property for each user will be extracted and included in the `dim_ga4__users` model by configuring the `user_properties` variable in your project as follows:
-
-```
-vars:
-  ga4:
-    user_properties:
-      - user_property_name: "membership_level"
-        value_type: "int_value"
-      - user_property_name: "account_status"
-        value_type: "string_value"
-```
-
-### Derived User Properties
-
-Derived user properties are different from "User Properties" in that they are derived from event parameters. This provides additional flexibility in allowing users to turn any event parameter into a user property. 
-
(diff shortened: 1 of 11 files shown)


dgitis:
The tests don't fail on my local computer.

adamribaudo-velir (approved):

adamribaudo-velir:
@dgitis This is the same error that we're seeing in PR #346. dbt-core doesn't understand what to do with a unit test when the model is disabled. 

Thanks for calling this out, @jice-lavocat if you enable the stg_ga4__users model, this error should go away. But that assumes you have a users table to stage.

jice-lavocat:
... I didn't disable this model.

In total, three unit tests are failing for me:
- test_base_to_stg_ga4__client_keys
- test_stg_ga4__client_keys_audiences
- test_stg_ga4__users_audiences

I'm not up to date on unit tests, but the test `test_base_to_stg_ga4__users` works, so I don't know what's different with the test `test_stg_ga4__users_audiences`.

dgitis:
Sorry about the late response. I saw this the morning that you posted and then totally forgot about it.

I suspect that the unit tests are failing because those modes aren't enabled. I tested whether the models work but not whether the the tests caused errors when the models aren't in use.

You should be able to disable the offending tests with the following block:

```
tests:
  staging:
    test_stg_ga4__users_audience:
      +enabled: false
    test_stg_ga4__client_keys_audiences:
      +enabled: false
    base:
      test_base_to_stg_ga4__client_keys:
        +enabled: false
```

I'm not sure if those resource paths are exactly correct. I haven't disabled tests this way but that should get you close.

I'll check disabling tests and update the documentation.

PR #366, opened 2025-05-13
Support ga_session_id v2

dgitis:
## Description & motivation

I'm not sure that we should officially support this yet so I'm leaving this as a draft for now.

I have one client where I'm seeing `string_value` `ga_session_id` values.

They appear in the following format: 's1747075570$o5$g1$t1747077281$j36$l0$h0'

Here's a breakdown of the data in this format:

-- s1747075570: Session ID
-- $o1: Session Count
-- $g0: Session Engaged
-- $t1747077281: Timestamp of first event in session
-- $j0: Countdown
-- $l0: Unknown
-- $h0: Enhanced Client ID (User ID?)

I've verified these changes on the client site where I observed this issue.

Is the pytest suite still broken? It was trying to connect to a project that it shouldn't have been connecting to and all tests errored.

We may also choose to merge this PR but not officially document the feature until we hear of others having the same problem. 

Regardless, I have chosen to not add update the readme file or add tests to this update for now.

## Checklist

### DIFF
--- a/macros/unnest_key.sql
+++ b/macros/unnest_key.sql
@@ -1,20 +1,51 @@
 -- Unnests a single key's value from an array. Use value_type = 'lower_string_value' to produce a lowercase version of the string value
+-- integrate_string_value_ga_session_id override:
+-- string_value ga_session_id in following format: 's1747075570$o5$g1$t1747077281$j36$l0$h0'
+-- s1747075570: Session ID
+-- $o1: Session Count
+-- $g0: Session Engaged
+-- $t1747077281: Timestamp of first event in session
+-- $j0: Countdown
+-- $l0: Unknown
+-- $h0: Enhanced Client ID (User ID?)
 
 {%- macro unnest_key(column_to_unnest, key_to_extract, value_type = "string_value", rename_column = "default") -%}
     {{ return(adapter.dispatch('unnest_key', 'ga4')(column_to_unnest, key_to_extract, value_type, rename_column)) }}
 {%- endmacro -%}
 
 {%- macro default__unnest_key(column_to_unnest, key_to_extract, value_type = "string_value", rename_column = "default") -%}
-    (select 
-        {% if value_type == "lower_string_value" %}
-            lower(value.string_value)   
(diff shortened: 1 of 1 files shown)


willbryant:
I am now getting those GS2 strings as the actual cookie values, but they definitely don't come through like that in the GA4 events sent by Google's own code, for me; they get extracted to session ID and count fields like normal.

But I did have to update our own server-side parsing code to do the same extraction so that events I send from our own server (via the Measurement Protocol) get integrated into the session correctly.

Is it possible your client had their own cookie capture/parsing code, that falls back to passing through the whole cookie value as the session ID?

dgitis:
@willbryant, it is definitely possible that it is an implementation issue.

The client uses a third-party tool for ecommerce tracking. They could be doing something. The fact that on some days `purchase` events are the only events affected definitely points to an issue with the integration.

However, the inconsistencies in how the data is passed points to a processing issue on Google's part.

Specifically, from May 01-May 05 (I haven't checked earlier), the `purchase` events are sent in a mix of `int_value` and `string_value`. This is what we would see if Google were rolling out a new Gtag library and some visitors were using an older cached version.

On May 06 and May 07, I had batch partitions where all events had `ga_session_id` in the `string_value` format.

Google later reprocessed these batch partitions so that all events except `purchase` events were in `int_value`.

The intraday continues to have `int_value` `ga_session_id` for all events. When Google processes the intraday tables, they get split into `string_value` purchase events and `int_value` other events.

I don't have a logical explanation for why this is so inconsistent. If it were an implementation issue affecting `purchase` events, I would expect the intraday tables to pass `string_value` session IDs just like the batch tables.

There is now a another person reporting a similar issue on Measure Slack so I'm thinking this is probably worth merging. It may be an implementation issue that's only surfaced with changes that Google has made to the Gtag library, but it's appearing less like it is an isolated issue.

dgitis:
Looking at the failing tests, none of them are related to the `unnest_key` macro so those issues are likely unrelated to any changes I'm suggesting here.

mtcarlone:
Hi - getting up to speed with the project. Please give me a little bit of time to go through this and get acquainted / review. Thanks in advance!

mtcarlone:
this will go out next week as a part of our Jan release!

PR #389, opened 2026-02-04
Create Macro encrypt_column.sql: PHI_MASK column

BindusekharGorintla:
Why Your encrypt_column Macro Is Useful
1. Reusability You don’t need to repeat hashing logic in every model. Example: {{ encrypt_column('email', 'sha256') }}
2. Consistency Ensures the same hashing method is applied everywhere. Reduces risk of mistakes (e.g., forgetting TO_HEX or CAST).
3. Flexibility You can switch hashing methods (md5, sha1, sha256) by passing a parameter. Makes it easy to adapt if your organization changes its security standards.

## Checklist

### DIFF
--- a/macros/encrypt_column.sql
+++ b/macros/encrypt_column.sql
@@ -1,0 +1,14 @@
+-- macros/encrypt_column.sql
+
+{% macro encrypt_column(column_name, method='sha256') %}
+    {%- if method == 'md5' -%}
+        TO_HEX(MD5(CAST({{ column_name }} AS STRING)))
+    {%- elif method == 'sha1' -%}
+        TO_HEX(SHA1(CAST({{ column_name }} AS STRING)))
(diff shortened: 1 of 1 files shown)


dgitis:
I think this would be more useful if it worked with the already existing key-gen columns. These are found in [stg_ga4__events](https://github.com/Velir/dbt-ga4/blob/main/models/staging/stg_ga4__events.sql).

Here's a tricky one as an example:
```
include_session_key as (
    select 
        *,
        to_base64(md5(CONCAT(client_key, CAST(session_id as STRING)))) as session_key
    from include_client_key
),
```

willbryant:
Encryption is not the same thing as hashing!

This is something you can do in your own project. If you want to be able to override the hashing in this project, it would be more useful to extract the existing hashing to a macro that you can then override.

Issue #320, opened 2024-04-29
Add specified user properties to event models


adamribaudo-velir:
We allow users to list out specific user properties which are then added to this model: https://github.com/Velir/dbt-ga4/blob/main/models/staging/stg_ga4__user_properties.sql

However, it would be useful and inexpensive to include these user properties in all event models. 

We perform a similar operation with custom event properties which are defined in the project.yml and then included in the event models with this code block:

` {% if var("default_custom_parameters", "none") != "none" %}
        {{ ga4.stage_custom_parameters( var("default_custom_parameters") )}}
      {% endif %}`

Which uses this macro: https://github.com/Velir/dbt-ga4/blob/main/macros/stage_custom_parameters.sql

A similar 'stag user properties' macro could be created to accomplish the same thing for user properties.

PR #396, opened 2026-05-28
issue320: adding stage_user_properties macro

valbelova:
## Description & motivation
https://github.com/Velir/dbt-ga4/issues/320

we want to enhance the [stg_ga4__event_page_view](https://github.com/Velir/dbt-ga4/blob/main/models/staging/events/stg_ga4__event_page_view.sql) model in the [velir/ga4 package](https://github.com/Velir/dbt-ga4) to support persisting an configurable set of user properties.

This feature would be an improvement over our current approach where we need to hard-code unnest(user_properties) calls in order to persist a user property at the event level 

## Checklist


valbelova:
@EvavW take a look

dgitis:
We already have this in several flavors. 

[Here are the docs](https://github.com/Velir/dbt-ga4/tree/main#dbt-ga4-original-user-table-settings).

valbelova:
> We already have this in several flavors.
> 
> [Here are the docs](https://github.com/Velir/dbt-ga4/tree/main#dbt-ga4-original-user-table-settings).

Hey, @dgitis thanks for flagging that! You're right that the existing user_properties var already handles extracting user
  properties into dim_ga4__users. What I'm building is slightly different. it adds user property values as columns
  directly on stg_ga4__event_page_view rows, so they're available at the event level rather than just in the user-level
  table.

  To avoid any unintended side effects for people who already have user_properties configured, I've updated the
  implementation to use a separate opt-in variable called page_view_user_properties instead. That way it stays independent
   from the existing behavior.

  Let me know if you have any questions or feedback!

dgitis:
I appreciate your enthusiasm contributing to the package but I'm of the opinion that we don't want to add this directly to the package because it adds complexity without adding anything new. 

The package is designed for users to join the tables later but it's also intended for people to override the models. I do this frequently for clients but only rarely do these customizations result in a PR. 

You can disable models using the `models` config path in your `dbt_project.yml` to the package model path and setting it to `enabled: false`. Then, copy the model code from the package to your project and customize as desired.

valbelova:
@dgitis Thanks for the feedback. I took a different approach that I think addresses your concern. Rather than adding it to a specific event model, I've added support for a default_user_properties variable directly in stg_ga4__events. This makes user properties available across all downstream event models generically, rather than as a one-off customization. Happy to hear your thoughts.

dgitis:
I also think I might have been a little hasty in my objection. The user_properties values in the event models are different from those in the user models. 

I would hope that the user models are a better place for them, but that's not really a good assumption.

I will review this more thoroughly when I have a few moments. Thanks again.

valbelova:
@dgitis Hello, have you had a chance to review this?

valbelova:
@dgitis @mtcarlone

dgitis:
Okay, so I looked at the PR and had a think about it.

I don't have a problem with the PR. My problem is that we've allowed too many user features to trickle in. The solution to this is to split the package with a 7.x version that works with the newer tables that Google has introduced, like the new user tables, and breaks compatibility with all of the older stuff in the package that we developed when Google didn't export quite as much as they do today.

However, Google does include the user properties in the events export so we should probably support their use.

So, the arguments for and against are primarily philosophical. I don't have a strong enough opinion either way.

Finally, I don't currently have a site that I can test the user properties on. I could set up my test site with user properties, but it points to another problem: these changes lack tests. 

Can you add a unit test, like the one at `/unit_tests/test_stg_ga4__derived_user_properties.py`? And then edit the README file to document the new setting?

valbelova:
@dgitis Sounds good, thanks! Added the unit test and updated the README. Let me know if there is anything I can add or change.

dgitis:
Obviously, there's an issue with the unit tests that I'm going to fix in a separate PR. 

I ran the unit test on my machine and it passed, thank you.

Can you update the tests to add a mock row that tests int_value (changing BBB to that is fine) and add a fixture that tests `rename_to`?

valbelova:
@dgitis added!

dgitis:
@mtcarlone, this looks ready to me.

mtcarlone (approved):
lgtm!

one nitpick - it may be clearer to add a leading comma in the `stg_ga4__events` model itself

```sql
...
  select
      *
      , {{ ga4. ... }}
```

and have the macro use `{% if not loop.last %}, {% endif %}`.

Not a blocker to releasing this, but I had to refer back to the macro to make sure we had all the commas in place.

mtcarlone:
This will go out September 8, end of day.

